PR #6, opened 2016-10-10
Adjusted for npm v3 rules directory

Blackbaud-BobbyEarl:
@Blackbaud-JonathanBell can you confirm this is what you had to change?


Blackbaud-BobbyEarl:
Jon confirmed offline this is indeed the change needed.

PR #22, opened 2016-11-11
Added version requirements

Blackbaud-BobbyEarl:
Using both `.nvmrc` and engines entry in `package.json`.

Part of https://github.com/blackbaud/sky-pages-cli/issues/31


package.json
```diff
@@ -14,6 +14,9 @@
     "type": "git",
     "url": "git+https://github.com/blackbaud/sky-pages-out-skyux2.git"
   },
+  "engines": {
+    "node": ">= 6.0.0"
```
Blackbaud-PaulCrowder: We might want to do >= 6.3.0 since I think there were some changes related to debugging in that version we may be relying on.

PR #31, opened 2016-12-08
Conditionally setting skyuxPathAlias to handle weird merging

Blackbaud-BobbyEarl:


test/cli-build.spec.js
```diff
@@ -135,6 +137,7 @@ describe('cli build', () => {
           expect(removeSpy).toHaveBeenCalledWith(
             skyPagesConfigUtil.spaPathTemp()
           );
+          expect(passedConfig.skyuxPathAlias).toBeUndefined();
```
Blackbaud-PaulCrowder: Does `toBeUndefined` pass if the property is there but has the value of `undefined`?  Maybe this should be `expect(passedConfig.hasOwnProperty('skyuxPathAlias')).toBe(false);` instead since the original bug was that the property was present but undefined.

test/cli-build.spec.js
```diff
@@ -135,6 +137,7 @@ describe('cli build', () => {
...
```
Blackbaud-PaulCrowder: Should we also have a test that checks for the presence of `import { SkyModule } from 'whatever'` in the resulting module file?  That way we don't have to worry about how we actually implemented overriding default behavior because we're ensuring the output itself is correct.

test/cli-build.spec.js
Blackbaud-BobbyEarl: Agreed.

PR #35, opened 2017-01-05
Upgraded to SKY UX 2 beta; renamed SKY Pages to SKY UX Builder

Blackbaud-PaulCrowder:


Blackbaud-PaulCrowder:
In the interest of not ripping up too much stuff I left any internal variables, file names, etc. that reference SKY Pages as-is.  I think we can continue to use `skyux` as the config property name since it'll match the repo name.

PR #64, opened 2017-02-07
Updated rxjs version due to issue with extending http class.

Blackbaud-SteveBrush:
- Open `skyuxconfig.json` and add `"auth": true`. Build should fail.


Blackbaud-BobbyEarl:
Hey @Blackbaud-SteveBrush.  A few of our dependencies were allowing newer versions of angular as well.  I'm taking this opportunity to also try and fix those, including what you've done here, in #67.

PR #65, opened 2017-02-07
[HOLD] Added Help Client to bootstrapper

Blackbaud-SteveBrush:
- jQuery no longer a dependency for BBHelp
- Adjusted the `skyuxconfig.json` file to include a "packages" attribute, to store all three module's configuration. (It seemed easier to implement, and will flex well in the future if we need to add more modules.)
```
"packages": {
  "auth": true,
  "help": {},
  "omnibar": {}
}
```


package.json
```diff
@@ -39,6 +39,7 @@
     "@angular/platform-browser-dynamic": "2.4.2",
     "@angular/router": "3.4.2",
     "@blackbaud/auth-client": "blackbaud/auth-client",
+    "@blackbaud/help-client": "blackbaud/help-client",
```
Blackbaud-SteveBrush: I'm unsure if this is the correct way to implement a namespaced module...

Blackbaud-BobbyEarl: This is the correct way until we being publishing to npm.

runtime/bootstrapper.ts
```diff
@@ -1,29 +1,32 @@
...
+      }
+
+      // BBHelp init
+      if (SkyAppBootstrapper.bootstrapConfig.help) {
+        promises.push(BBHelp.load(SkyAppBootstrapper.bootstrapConfig.help));
       }
     }
 
-    return Promise.resolve();
+    return Promise.all(promises);
```
Blackbaud-BobbyEarl: Is the processBootstrapConfig resolved if auth, omnibar, and help are all false/not configured?

lib/sky-pages-bootstrap-generator.js
```diff
@@ -4,16 +4,11 @@
...
-SkyAppBootstrapper.bootstrapConfig = {
-  omnibar: ${JSON.stringify(skyPagesConfig.omnibar)},
-  auth: ${JSON.stringify(skyPagesConfig.auth)}
-};
-
-`;
+  if (skyPagesConfig.packages) {
+    bootstrap = `
+      import { SkyAppBootstrapper } from '${skyPagesConfig.runtimeAlias}';
+      SkyAppBootstrapper.bootstrapConfig = ${JSON.stringify(skyPagesConfig.packages)};
```
Blackbaud-BobbyEarl: I can appreciate the efficiency by not having to specify auth, help, and omnibar in two places (here and `runtime/bootstrap.ts`, which has to stay).  

Doing this means we would have to instruct our current consumers to update their `skyuxconfig.json`, which I'm not sure is worth the efficiencies gained.  Generally speaking, we've found being very explicit in what options / configuration we allow has worked best.

Blackbaud-SteveBrush: Fair enough. I'll revert to a more deliberate approach (your tests also would have to be rewritten to accommodate, so probably not worth the effort).

PR #66, opened 2017-02-08
Added Help client to bootstrapper

Blackbaud-SteveBrush:
screen shot 2017-02-08 at 1 36 17 pm


runtime/bootstrapper.ts
```diff
@@ -1,29 +1,32 @@
...
+      }
+
+      // BBHelp init
+      if (SkyAppBootstrapper.bootstrapConfig.help) {
+        promises.push(BBHelp.load(SkyAppBootstrapper.bootstrapConfig.help));
       }
     }
 
-    return Promise.resolve();
+    return Promise.all(promises);
```
Blackbaud-BobbyEarl: I started to ask if this would resolve in the case of auth, omnibar, and help being set as false, which means `bootstrapConfig` is never set.  

Looking at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all says "If an empty array is passed, then this method resolves immediately.", so question answered.

PR #92, opened 2017-03-12
Added support for config flag

Blackbaud-BobbyEarl:


Blackbaud-BobbyEarl:
This is a first pass at some of the of the initial work required to support a different configuration file(s).

@Blackbaud-TerryHelems and I talked about this PR, and he brought up some good points, that there may be more value in implicitly reading in the configuration file, something similar to how dotnet core works.

Via the build definition, we could expose the current environment.  If a config filename matches the given environment, we automatically consume it.  This could also apply while working locally with `skyux serve`.

Consumers can already execute custom tasks via npm `preinstall` and `postinstall` hooks, so exposing the environment could allow them to branch on that logic as well.

@blackbaud-PeteLukich and @Blackbaud-MitchellThomas were interested, so I'm curious if you have any feedback as well.

Blackbaud-BobbyEarl:
Updated this PR based on some initial feedback.  A few changes:

- Still supporting the `--config` and `-c` flags, but only for a single file.
- Added support for an `extends` property in any config file to support multiple, merged files.
- Reading `SKYUX_ENV` environment variable.  If it and a corresponding `skyuxconfig.SKYUX_ENV.json` file exists, use it.

Blackbaud-BobbyEarl:
Okay, another revision to this PR:

- Removed config flag
- By default, if it exists, `skyuxconfig.json` is still read from the root.
- Instead of relying on environment variables, put a focus on the current command / context being run.  If a matching `skyuxconfig.[COMMAND].json` is found, automatically merge that config file in.  Again, automatically extending `skyuxconfig.json` if it exists.

Blackbaud-MitchellThomas:
This looks clean, @Blackbaud-BobbyEarl. I like the model, and if I can access this config inside the SPA, this would do exactly what my team needs for configuring the backend url.

Blackbaud-BobbyEarl:
Thanks @Blackbaud-MitchellThomas!  I believe this is ready for your review @Blackbaud-PaulCrowder.

Blackbaud-PaulCrowder:
@Blackbaud-BobbyEarl @Blackbaud-MitchellThomas mentioned that he would need to access the config file inside the SPA.  Do we have a clean way for consumers to do this?  Preferably we would implement the equivalent of "app settings" so that connection strings can be specified in the config file itself and then just expose those settings in a first-class way.

Blackbaud-BobbyEarl:
@Blackbaud-PaulCrowder  Good question. Currently, user's could do something like the following in `skyux serve`.  It will fail in `skyux build` and their IDE will not like it unless we added `export let SkyPagesProvider: any;` to the `src/app/sky-pages.module.ts` file.

```
import { Component, Inject } from '@angular/core';
import { SkyPagesProvider } from '@blackbaud/skyux-builder/src/app/sky-pages.module';

@Component({
  selector: 'my-home',
})
export class HomeComponent {
  constructor(@Inject(SkyPagesProvider) SKY_PAGES: any) {
    console.log(SKY_PAGES);
  }
}
```

So, short answer is no, they cannot.  In my `provide-envid` branch (ignore most of it since it's in a really bad state), I attempted to make it better by doing a couple things.

1.  Creating the SkyPagesConfig type so we can be more explicit around our model.  This would have caught the bug solved in #90.  We could stick with `any` if you think this may be problematic.
2.  Defining the provider, interface, and variable in  [runtime/sky-pages.config.ts](https://github.com/blackbaud/skyux-builder/blob/provide-envid/runtime/sky-pages.config.ts).  This make the IDE, `skyux serve` and `skyux build` (with/without AOT) happy.

This would allow a similar component to look like the following, assuming we added `appSettings` to the `SkyPagesConfig` interface.

```
import { SkyAuthHttp, SkyPagesConfig } from '@blackbaud/skyux-builder/runtime';

@Component({
  selector: 'my-comp',
  templateUrl: './my.component.html'
})
export class MyComponent {
  constructor(@inject SKY_PAGES: SkyPagesConfig) {
    SkyAuthHttp.get(SKY_PAGES.appSettings.backendUrl + '/endpoint');
  }
}
```

Are you okay with that approach?  If so, I can bring them into this branch, or do my original plan, which was to include those in a new branch after this one is merged (not really dependent on it I guess).

Blackbaud-PaulCrowder:
Let's go ahead and merge this as-is and then talk about how best to implement app settings later.

PR #96, opened 2017-03-21
Allowing envid + svcid to be passed through

Blackbaud-BobbyEarl:
Tackling this in more manageable chunks.  First part (currently most-needed) to solve https://github.com/blackbaud/skyux2/issues/415

### DIFF
--- a/config/webpack/serve.webpack.config.js
+++ b/config/webpack/serve.webpack.config.js
@@ -16,6 +16,32 @@ const hostUtils = require('../../utils/host-utils');
 const moduleLoader = skyPagesConfigUtil.outPath('loader', 'sky-pages-module');
 
 /**
+ * Returns the querystring base for parameters allowed to be passed through.
+ * @name getQueryStringFromArgv
+ * @param {Object} argv
+ * @returns {string}
+ */
+function getQueryStringFromArgv(argv) {
+  const allowed = [
+    'envid',
+    'svcid'
+  ];
+
+  let found = [];
+  allowed.forEach(param => {
+    if (argv[param]) {
+      found.push(`${param}=${argv[param]}`);
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


test/config-webpack-serve.spec.js
```diff
@@ -280,4 +299,29 @@ describe('config webpack serve', () => {
...
+    expect(openCalledWith).toContain(`?svcid=asdf`);
+  });
+
+  it('should pass through envid and svcid, but not junk from the command line', () => {
+    argv.envid = 'asdf1';
+    argv.svcid = 'asdf2';
+    argv.myid = 'asdf3';
+
+    bindToDone();
+    expect(openCalledWith).toContain(`envid=asdf1`);
```
Blackbaud-PaulCrowder: Since you're testing with multiple query parameters here it might be helpful to write a test that validates whether they all form a valid query string (question mark at the beginning, ampersands between them, etc.)

config/webpack/serve.webpack.config.js
```diff
@@ -16,6 +16,32 @@ const hostUtils = require('../../utils/host-utils');
...
+function getQueryStringFromArgv(argv) {
+  const allowed = [
+    'envid',
+    'svcid'
+  ];
+
+  let found = [];
+  allowed.forEach(param => {
+    if (argv[param]) {
+      found.push(`${param}=${argv[param]}`);
```
Blackbaud-PaulCrowder: Any time you build a URL from user input you should pass it through the `encodeURIComponent()` method.  A good test for this would be to specify one of the values with an ampersand or question mark and make sure it gets escaped.

Blackbaud-BobbyEarl:
Incorporated your feedback @Blackbaud-PaulCrowder

PR #98, opened 2017-03-22
Adding envid and svcid to SkyAuthHttp

Blackbaud-BobbyEarl:


Blackbaud-BobbyEarl:
@Blackbaud-PaulCrowder Any suggestions or something blatant you see I may be missing regarding code coverage?  

In the [test.webpack.config.js](https://github.com/blackbaud/skyux-builder/blob/master/config/webpack/test.webpack.config.js#L117), if I override the `istanbul-instrument-loader` to look at my `runtimePath` instead of the default `srcPath`, I always get "unexpected token, line 1" errors.

Blackbaud-BobbyEarl:
I was able to solve the `unexpected token` error mentioned above.

PR #102, opened 2017-04-03
Resolving error with resolving loader.

blackbaud-brandonhare:
Webpack 2 requires loaders to use the `-loader` suffix. Without it, an error is thrown.  https://webpack.js.org/guides/migrating/#automatic-loader-module-name-extension-removed

### DIFF
--- a/utils/sky-style-loader.js
+++ b/utils/sky-style-loader.js
@@ -2,7 +2,7 @@
 
 'use strict';
 
-require('style!@blackbaud/skyux/dist/css/sky.css');
+require('style-loader!@blackbaud/skyux/dist/css/sky.css');
 
 var FontFaceObserver = require('fontfaceobserver');
 


Blackbaud-BobbyEarl:
Hey @blackbaud-brandonhare, I definitely see what you've proposed in the migration guide.  Just wondering if you're actually seeing an error anywhere?.

blackbaud-brandonhare:
I saw the error when I ran skyux test in my project. I had similar errors in our stache stuff that I fixed.

Sent from my iPhone

On Apr 3, 2017, at 5:55 PM, Bobby Earl <notifications@github.com<mailto:notifications@github.com>> wrote:


Hey @blackbaud-brandonhare<https://github.com/blackbaud-brandonhare>, I definitely see what you've proposed in the migration guide. Just wondering if you're actually seeing an error anywhere?.

Blackbaud-BobbyEarl:
Hey @blackbaud-brandonhare, after talking and digging in a bit more, the error only showed up for you as you were experimenting with a newer version of Webpack globally.

When we update the version of Webpack in builder, we'll definitely need to make this adujstment.  There are actually a couple additional places where we'll need to include the `-loader` suffix if you're interested in updating all of them.  

Otherwise I'm going to close it until we update to the version of Webpack that will display this error.

PR #112, opened 2017-04-13
Updated package.json and CHANGELOG.md

Blackbaud-BobbyEarl:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 1.0.0-beta.15 (2017-03-22)
+
+- Images! Images stored in `src/assets` and referenced via `~/assets/` in all HTML/SCSS files are now bundled.
```
blackbaud-johnly: Why do we have "Images!" at the start? ... Usually we start these entries with a past tense verb. I'd flip this to "Bundled images stored in..." and delete "are now bundled" at the end.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+- Added `preload` and `postload` plugin hooks.  Beginning initial work for a lifecycle hooks. (Thanks @Blackbaud-SteveBrush!)
```
blackbaud-johnly: I'd change "Beginning" to "Started" to be consistent with the past tense verb in the previous sentence. ... Also, I think we should delete "a" before "lifecycle hooks" since it is plural. "Either that or change "hooks" to "hook.")

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+- Automatically including `envid` and `svcid` in `SkyAuthHttp` requests.
```
blackbaud-johnly: The verbs that start these are usually past tense. I'd change "including" to "included."

CHANGELOG.md
Blackbaud-BobbyEarl: Because I was excited about it.  :-)

PR #115, opened 2017-04-17
Added SkyAppConfig to SkyAppTestModule's list of providers

Blackbaud-SteveBrush:
Fixes [Issue #595](https://github.com/blackbaud/skyux2/issues/595)

### DIFF
--- a/lib/sky-pages-module-generator.js
+++ b/lib/sky-pages-module-generator.js
@@ -51,6 +51,11 @@ function getSource(skyAppConfig) {
     'AppExtrasModule'
   ];
 
+  runtimeProviders.push(`{
+    provide: SkyAppConfig,
+    useValue: ${skyAppConfigAsString}
+  }`);
+
   if (skyAppConfig.runtime.includeRouteModule) {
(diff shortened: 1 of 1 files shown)


lib/sky-pages-module-generator.js
```diff
@@ -52,19 +52,17 @@ function getSource(skyAppConfig) {
...
-
     runtimeModuleImports.push('routing');
-
     runtimeProviders.push('appRoutingProviders');
+  } else {
+    /*
+      Import the regular RouterModule so that tested components can reference things
+      like routerLink
+    */
     runtimeProviders.push(`{
```
Blackbaud-BobbyEarl: I believe this line may need to always be included.

Blackbaud-PatrickOFriel: Agreed, just place it outside of the if/else

Blackbaud-BobbyEarl: I would probably suggest adding it a little higher up at the `runtimeProviders` declaration.

PR #153, opened 2017-05-16
added a replace string to remove BOM from json file before parse

Blackbaud-BrandonJones:
Currently if editing the skyuxconfig.json file in Visual Studio, it addes a Byte Order Mark to the file, which causes the json parse to explode.  

https://github.com/nodejs/node-v0.x-archive/issues/1918#issuecomment-2480359

This is a work around as readFileSync doesn't strip out BOM by default. 
Related to:
https://github.com/blackbaud/skyux2/issues/700

### DIFF
--- a/config/sky-pages/sky-pages.config.js
+++ b/config/sky-pages/sky-pages.config.js
@@ -19,7 +19,8 @@ function resolve(root, args) {
 }
 
 function readConfig(file) {
-  return JSON.parse(fs.readFileSync(file, 'utf8'));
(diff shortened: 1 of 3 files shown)


Blackbaud-PaulCrowder:
Does using `readJsonAsync` from `fs-extras` also work?  `fs-extras` uses the `jsonfile` package to read JSON documents which [strips the BOM](https://github.com/jprichardson/node-jsonfile/blob/master/index.js#L32) before parsing the document.

PR #166, opened 2017-05-25
Fixed FOUC

Blackbaud-SteveBrush:
Addresses issue [707](https://github.com/blackbaud/skyux2/issues/707) and [157](https://github.com/blackbaud/stache2/issues/157)


tsconfig.json
```diff
@@ -13,9 +13,8 @@
       "dom",
       "es6"
     ],
-    "types": [
-      "jasmine",
-      "node"
+    "typeRoots": [
```
Blackbaud-BobbyEarl: Any insight into what this change does?

src/app/app.component.html
```diff
@@ -1 +1,4 @@
-<router-outlet></router-outlet>
+<div class="sky-pages-cloak" [ngClass]="{ 'visible': isReady }">
```
Blackbaud-BobbyEarl: How about `skyux-spa-cloak`?

runtime/style-loader.ts
```diff
@@ -0,0 +1,35 @@
...
+        oswald.load(undefined, SkyAppStyleLoader.LOAD_TIMEOUT)
+      ])
+      .then(() => {
+        this.isLoaded = true;
+      })
+      .catch((error) => {
+        // Errors loading the font should not stop the page from rendering.
+        // Passing the error along in case the client wants to do something with it.
+        return Promise.resolve({
+          error: error
```
Blackbaud-BobbyEarl: Is `error` ever used?  I agree it makes sense to have the app still load in the event of an error, but maybe we could at least console.log it, or even add something to the template?

```
<div *ngIf="error">
  There was an error loading the requested fonts.  The application is still available.
</div>
```

tsconfig.json
Blackbaud-SteveBrush: It allows tsc to find types for `fontfaceobserver`. I could be deliberate and add `"fontfaceobserver"` to the list, if that's preferred.

src/app/app.component.html
Blackbaud-SteveBrush: Yeah, I can do that. I was mimicking the `id` of the app's wrapping div.

runtime/style-loader.ts
Blackbaud-SteveBrush: Well, the "error" will still fire if the font is taking too long to load (so, it might still exist, but is taking longer than 3 seconds to load). We can console log it if that helps, but thought it would be helpful to provide the error to the consuming service, if they wanted to check for it. Otherwise, there's no way for the consumer to "know" that an error occurred, because it will always be resolved:
https://github.com/blackbaud/skyux-builder/pull/166/files#diff-df8c52ab04ab7a3e6b86e304a41fd060R30

Blackbaud-BobbyEarl: I believe by putting my comment in the wrong spot it's thrown us off.  I'm going to comment in the right place.

src/app/app.component.ts
```diff
@@ -70,12 +71,19 @@ function fixUpNav(nav: any, baseUrl: string) {
...
+
   constructor(
     private router: Router,
     private windowRef: SkyAppWindowRef,
     private config: SkyAppConfig,
+    private styleLoader: SkyAppStyleLoader,
     @Optional() private searchProvider?: SkyAppSearchResultsProvider
-  ) { }
+  ) {
+    this.styleLoader.loadStyles().then(() => {
```
Blackbaud-BobbyEarl: From our previous conversation, my suggestion was something like this:

```
public hasError = false;
this.styleLoader.loadStyles.then(error => {
  this.isReady = true;
  if (error) {
    console.log(error);
    this.hasError = true;
  }
});
```

And then in `app.component.html`, something like:
```
<div *ngIf="hasError">
  There was an error loading the requested fonts.  The application is still available.
</div>
```

src/app/app.component.html
Blackbaud-BobbyEarl: Ahh I see.  Ignore my comment.  :-)

src/app/app.component.ts
Blackbaud-SteveBrush: Ah, that makes sense. Do we want to show an error on the front-end, though? What if we put it in the red banner?

Blackbaud-BobbyEarl: The red banner is a implementation in host only, not in builder.  We've reserved that to indicate when a site is being served locally.

I'm for visually showing something, but perhaps a console.log is sufficient.

Perhaps @Blackbaud-PaulCrowder has an opinion?

tsconfig.json
Blackbaud-BobbyEarl: Ok, looking at this a little more.  Reading over the "@types, typeRoots and types" section of https://www.typescriptlang.org/docs/handbook/tsconfig-json.html, it seems you've put back in what is the default behavior.

Since you're importing `fontFaceObjserver` using the module syntax, I would've expected this sentence to be true.  

"Keep in mind that automatic inclusion is only important if you’re using files with global declarations (as opposed to files declared as modules). If you use an import "foo" statement, for instance, TypeScript may still look through node_modules & node_modules/@types folders to find the foo package."

src/app/app.component.ts
Blackbaud-PaulCrowder: I'd prefer to gracefully fail and not show an error message.  Writing something to `console.log` should be sufficient.

src/app/app.component.html
Blackbaud-PaulCrowder: I'd rather see an `ngClass` attribute that adds a CSS class when `!isReady` so that it's removed when `isReady` is true.  That way you only need one CSS class instead of two.  As for the name, I'd use something like `skyux-app-loading`.

tsconfig.json
Blackbaud-SteveBrush: Removing this change from this branch as it doesn't seem to be throwing the same error anymore. I guess I fixed it inadvertently.

runtime/style-loader.ts
```diff
@@ -0,0 +1,38 @@
...
+        // when the font is loaded unless a known character with a different width
+        // than the default is not specified.
+        fontAwesome.load('\uf0fc', SkyAppStyleLoader.LOAD_TIMEOUT),
+        openSans.load(undefined, SkyAppStyleLoader.LOAD_TIMEOUT),
+        oswald.load(undefined, SkyAppStyleLoader.LOAD_TIMEOUT)
+      ])
+      .then(() => {
+        this.isLoaded = true;
+        return Promise.resolve({
+          status: true
```
Blackbaud-BobbyEarl: Hey @Blackbaud-SteveBrush is `status` used somewhere else?

Blackbaud-SteveBrush: It's just another way to detect if the resolution was a success (`result.error`, `result.status`). It's either this or an empty resolve (we need a resolve either way). What do you think?

Blackbaud-SteveBrush: Maybe, `return Promise.resolve({ error: undefined })`? The issue comes when you need to check if `result.error` exists. If there isn't an error, `result.error` throws an error because `result` is undefined.

```
this.styleLoader.loadStyles().then((result: any) => {
  if (result.error) { // <-- throws error
  }
});
```

Blackbaud-BobbyEarl: Thoughts on just making result optional?
```
this.styleLoader.loadStyles().then((result?: any) => {
  if (result && result.error) {
  }
});
```

Then I think you could also remove that unnecessary return Promise.

```
    return Promise
      .all([
        // Specify a character for FontAwesome since some browsers will fail to detect
        // when the font is loaded unless a known character with a different width
        // than the default is not specified.
        fontAwesome.load('\uf0fc', SkyAppStyleLoader.LOAD_TIMEOUT),
        openSans.load(undefined, SkyAppStyleLoader.LOAD_TIMEOUT),
        oswald.load(undefined, SkyAppStyleLoader.LOAD_TIMEOUT)
      ])
      .then(() => {
        this.isLoaded = true;
      })
      .catch((error) => {
        // Errors loading the font should not stop the page from rendering.
        // Passing the error along in case the client wants to do something with it.
        return Promise.resolve({
          error: error
        });
      });
```

PR #167, opened 2017-06-01
Allowing browser flag in serve

Blackbaud-BobbyEarl:
Tested in OSX and Windows against `chrome`, `firefox`, `safari`, `iexplore`, and `edge`.  

A few notes:
- `edge` has to be a hard-coded use-case since they only expose a protocol `microsoft-edge:` instead of an executable.
- This only updates the `skyux serve` command.  Support could also be added for the `skyux test`, `skyux watch`, and `skyux e2e` commands.

Fixes https://github.com/blackbaud/skyux2/issues/402

### DIFF
--- a/config/webpack/serve.webpack.config.js
+++ b/config/webpack/serve.webpack.config.js
@@ -51,6 +51,11 @@ function getWebpackConfig(argv, skyPagesConfig) {
    * @name WebpackPluginDone
    */
   function WebpackPluginDone() {
+    const shorthand = {
+      l: 'launch',
+      b: 'browser'
+    };
+
     let launched = false;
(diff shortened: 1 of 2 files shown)


Blackbaud-BobbyEarl:
Good suggestion @Blackbaud-PaulCrowder.  Incorporated.

Blackbaud-BobbyEarl:
Just a heads up @blackbaud-johnly, this is something we'll want to document.  I started to write it, but saw the critiques already flowing, so I'll just save us both time and ask that you do it.  :-)

blackbaud-johnly:
Thanks @Blackbaud-BobbyEarl . I already had a note to ask you about this next week.

PR #168, opened 2017-06-02
Add support for route guards

blackbaud-brandonstirnaman:
Adds support for canActivate/canDeactivate route guards as requested in blackbaud/skyux2#423.

I want to look at refactoring the way route generation works to use childRoutes and all for canActivateChild as well but this is a solid first step to unblock some consumers.

### DIFF
--- a/lib/sky-pages-route-generator.js
+++ b/lib/sky-pages-route-generator.js
@@ -3,6 +3,7 @@
 
 const glob = require('glob');
 const path = require('path');
+const fs = require('fs');
 
 function indent(count) {
   return '  '.repeat(count);
@@ -53,6 +54,11 @@ function parseFileIntoEntity(skyAppConfig, file, index) {
 
   let routePath = [];
   let routeParams = [];
+  let parsedPath = path.parse(file);
+
+  // Make no assumptions on extension used to create route, just remove
+  // it and append .guard.ts (ex: index.html -> index.guard.ts)
+  let guardPath = path.join(parsedPath.dir, `${parsedPath.name}.guard.ts`);
 
   // Removes srcPath + filename
   // glob always uses '/' for path separator!
@@ -85,7 +91,8 @@ function parseFileIntoEntity(skyAppConfig, file, index) {
     componentName: componentName,
     componentDefinition: componentDefinition,
     routePath: routePath.join('/'),
-    routeParams: routeParams
+    routeParams: routeParams,
+    guardPath: fs.existsSync(guardPath) ? guardPath : undefined
   };
 }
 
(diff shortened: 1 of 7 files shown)


blackbaud-brandonstirnaman:
Don't merge this yet, testing some stuff related to deep nested routes.

blackbaud-brandonstirnaman:
Nested routes work as expected. Additionally after talking with @Blackbaud-PaulCrowder I'm going to make some changes so that guard creators don't have to `export default`.

blackbaud-brandonstirnaman:
With latest commit on this PR, it now supports using both named exports and a default export. Ready for review @Blackbaud-BobbyEarl.

lib/sky-pages-route-generator.js
```diff
@@ -114,11 +129,30 @@ function generateDefinitions(routes) {
 function generateDeclarations(routes) {
   const p = indent(1);
   const declarations = routes
-    .map(r => `${p}{ path: '${r.routePath}', component: ${r.componentName} }`)
+    .map(r => {
+      let guard = r.guardPath && r.guardName ? `require('${r.guardPath}').${r.guardName}` : '';
```
Blackbaud-PaulCrowder: Was there not a way to build a list of classes to import like we do in the code where we detect components and add them to the declarations of the NgModule?  I'd like to use that same pattern if at all possible.  Same for the `generateProviders` method on line 148.

test/sky-pages-route-generator.spec.js
```diff
@@ -104,4 +104,70 @@ describe('SKY UX Builder route generator', () => {
...
+    spyOn(fs, 'existsSync').and.returnValue(true);
+
+    let routes = generator.getRoutes({
+      runtime: {
+        srcPath: 'my-custom-src/',
+        routesPattern: 'my-custom-pattern',
+      }
+    });
+
+    expect(routes.declarations).toContain(
```
Blackbaud-PaulCrowder: This is still checking for `.default` from your previous implementation.  Should this be `.Guard` instead?  There are other instances later in this file, so be sure to address them as well.

blackbaud-brandonstirnaman: Nope. Its a separate test to verify that `.default` still works. Basically both paths are explicitly supported now. The test above uses the `.Guard` approach.

lib/sky-pages-route-generator.js
blackbaud-brandonstirnaman: Ok. gimme a few minutes.. be ready to review again soon! :)

test/sky-pages-route-generator.spec.js
blackbaud-brandonstirnaman: Removing support for `export default` entirely now as part of refactoring to use `import` instead of `require`

PR #170, opened 2017-06-06
Accessibility testing in e2e

Blackbaud-MattGregg:
Adding 2 new dependencies - axe-core & axe-webdriverjs 

New method to call in e2e tests as necessary in flow, called  - 
it('should pass all accessibility rules', (done) => {
    A11yTest.run().then(failures => {
      expect(failures).toBe("0 failures");
      done();
    });
  });

All aXe default rules are on by default. Can be turned off in project's skyuxconfig.json AppSettings property. Will be submitting pull request to SKYUX-template project with these all turned off.
Example - 
"appSettings":
    { "a11y": {
        "rules": {
          "area-alt": { "enabled": false },
          "audio-caption": { "enabled": false },....

Accessibility aXe failures error to console as in this example ----

Accessibility failure(s) found for: https://host.nxt.blackbaud.com/a11y-tests/

1 element failed 'label' rule: Form elements must have labels 
  Location: body > sky-pages-app > ng-component > my-home > input
  <input id="state" name="state" type="text">
  Get help at: https://dequeuniversity.com/rules/axe/2.2/label?application=webdriverjs


runtime/testing/e2e/a11y.ts
```diff
@@ -0,0 +1,46 @@
+/*jslint node: true */
+'use strict';
+
+const AxeBuilder = require('axe-webdriverjs');
+import { browser } from 'protractor';
+
+export class A11yTest {
```
Blackbaud-BobbyEarl: I would suggest we at a minimum prepend `Sky`, so `SkyA11y`.  Even though, I'm wondering if there's a better name?  `SkyAxeTools` maybe?

runtime/testing/e2e/a11y.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  public static log(results): any {
+      //console.error(`${violations}`);
```
Blackbaud-BobbyEarl: You can remove this line.  :-)

runtime/testing/e2e/a11y.ts
```diff
@@ -0,0 +1,46 @@
...
+      var numResults = results.violations.length;
+
+      if (numResults > 0) {
+        console.error("Accessibility failure(s) found for: " + results.url + "\n");
+        results.violations.forEach(result => {
+          var label = result.nodes.length === 1 ? 'element' : 'elements';
+          var msg = result.nodes.reduce(function(msg, node) {
+              return `${msg}  Location: ${node.target[0]}
```
Blackbaud-BobbyEarl: There's an extra space before `Location`.

runtime/testing/e2e/a11y.ts
```diff
@@ -0,0 +1,46 @@
...
+'use strict';
+
+const AxeBuilder = require('axe-webdriverjs');
+import { browser } from 'protractor';
+
+export class A11yTest {
+
+  public static log(results): any {
+      //console.error(`${violations}`);
+      var numResults = results.violations.length;
```
Blackbaud-BobbyEarl: Use `const` instead of `var`.

runtime/testing/e2e/a11y.ts
```diff
@@ -0,0 +1,46 @@
...
+    return new Promise(resolve => {
+      AxeBuilder(browser.driver)
+        .options(options)
+        .analyze(results => {
+          const violations = results.violations.length;
+          if (violations) {
+            A11yTest.log(results);
+          }
+          var fail = results.violations.length === 1 ? ' failure' : ' failures';
+          resolve(violations + fail);
```
Blackbaud-BobbyEarl: I think it's weird to return a string here instead of just the number of violations.

runtime/testing/e2e/a11y.ts
```diff
@@ -0,0 +1,46 @@
...
+  ${node.html}`;
+          }, "\n");
+          console.error(`${result.nodes.length} ${label} failed '${result.id}' rule: ${result.help} ${msg}
+  Get help at: ${result.helpUrl}
+`)
+        });
+      }
+  }
+
+  public static run(): Promise<Number> {
```
Blackbaud-BobbyEarl: Perhaps not in this initial scope, but we could add a parameter to the `run` method that would override / merge into the default options set in `skyuxconfig.json`.  That would allow specific tests to run against specific rules.

Blackbaud-BobbyEarl:
Hey @Blackbaud-MattGregg, did you have any questions about the changes I requested or need any help with anything?

runtime/testing/e2e/a11y.ts
Blackbaud-MattGregg: Sounds good. Going to go with SkyA11y. I intentionally didn't include "axe" anywhere as I thought this Sky class could down the road include other a11y testing libraries within it and teams wouldn't need to know those details and just have this one A11y call.

runtime/testing/e2e/a11y.ts
Blackbaud-MattGregg: Could be something to look at if teams need something like this as they start using. Maybe even to turn off tests for some reason. axe itself only run sthe tests it finds elements for that meet the criteria (i.e., only runs test for labels if there are labels in the doc/selector).

runtime/testing/e2e/a11y.ts
Blackbaud-MattGregg: this is intentional so each failure has a header line and then each failure instance found is indented under it.. 

```
1 element failed 'label' rule: Form elements must have labels 
  Location: body > sky-pages-app > ng-component > my-home > input
  <input id="state" name="state" type="text">
  Get help at: https://dequeuniversity.com/rules/axe/2.2/label?application=webdriverjs
1 element failed '"next rule"....
```

runtime/testing/e2e/a11y.ts
Blackbaud-MattGregg: Ok, ok. My brain wanted to make this a sentence in the output. - Expected '1 failure' to be '0 failures', but I'll update as maybe this isn't typical for what developers are used to seeing results.

Blackbaud-MattGregg:
@Blackbaud-BobbyEarl - no questions. I just wasn't following this repo so didn't get any emails. I made the updates I believe.

runtime/testing/e2e/a11y.ts
```diff
@@ -0,0 +1,42 @@
...
+    const config = JSON.parse(browser.params.skyPagesConfig);
+    const options = config.skyux.appSettings.a11y;
+
+    return new Promise(resolve => {
+      AxeBuilder(browser.driver)
+        .options(options)
+        .analyze(results => {
+          const violations = results.violations.length;
+          if (violations) {
```
Blackbaud-BobbyEarl: Since you're also checking for `results.violations.length` inside the `log()` method, you can remove the `if` statement here.  That, or remove it from the `log()` method, but I'd vote for the former.

Blackbaud-BobbyEarl:
Hey Matt... aside from my newest comment about double `if` statements, I think this looks good.

I don't know if we've talked about creating the corresponding unit tests.  These would exist in a new `a11y.spec.ts` file and contain multiple tests to validate the logic in your class.  Luckily, there's not a lot of code there so I would expect 3 - 5 tests to be able to cover everything.

Is that something you'd be comfortable contributing as well?  I'm happy to meet to get you started.

Blackbaud-SteveBrush:
Docs, FYI: http://accessibility.docs.blackbaudhosting.com/gettingstarted/implement_playbook/2-prepe2e/

Blackbaud-SteveBrush:
Closing this pull request in favor of https://github.com/blackbaud/skyux-builder/pull/250. There is a [Protractor plugin](https://github.com/angular/protractor-accessibility-plugin) that will handle the bulk of the logic for us. Also, the requirement for the feature is that accessibility checks execute automatically during `skyux e2e`.

PR #171, opened 2017-06-06
Fixed deps + test

Blackbaud-BobbyEarl:


runtime/auth-http.spec.ts
```diff
@@ -128,4 +128,14 @@ describe('SkyAuthHttp', () => {
     });
   });
 
+  /**
+   * PLEASE NOTE
+   * If this tests fails, it means you've changed the required parameters to the constructor.
+   * To successfully maintain backwards compatibility, make any new parameters optional.
```
Blackbaud-PaulCrowder: You might also want to specify that new parameters must also be added to the end of the constructor for the case where someone's calling the constructor by using `useFactory()`:

```
providers: [
  provide: SkyAuthHttp,
  useFactory: (a, b, c) => {
    return new SkyAuthHttp(a, b, SomethingElse);
  },
  deps: [a, b, c]
]
```

PR #173, opened 2017-06-06
Modernizing ProcessExitCode plugin

Blackbaud-BobbyEarl:
Fixes https://github.com/blackbaud/skyux2/issues/755


Blackbaud-BobbyEarl:
I don't believe my changes have introduced a reduction in code coverage.  Since I removed code that was previously covered, there is now a smaller total amount of code, but the amount uncovered stayed the same, meaning a smaller percentage of code is now covered.

PR #176, opened 2017-06-09
Child route support, root guard support

stirno:
PR implements child routing based on the design @Blackbaud-PaulCrowder and I discussed yesterday.

A directory prefixed with `#` such as `src/app/about/#child` would create a child-route underneath `/about` which can be affected by `canActivateChild` guards as well as used by controls that implement `router-outlet` in their template such as the future tab control (or currently microedge contrib tab control).

A directory without the prefix but underneath a child route such as `src/app/about/#child/top-level` will break out and become a top level route that cannot be affected by `canActivateChild` guards and will not participate in `router-outlet`.

All other routes will continue to be top-level. Routes continue to only be created if an `index.html` is provided.

Additionally now that we have child routing, I've also implemented a top-level route that makes it very simple to declare an `index.guard.ts` next to `src/app/index.html` and have a guard that affects being able to load *anything* within the SPA. This should be the preferred method for locking down SPAs in future.

Added several e2e tests to validate the functionality as well as unit tests for the route generation.

### DIFF
--- a/e2e/shared/common.js
+++ b/e2e/shared/common.js
@@ -212,6 +212,34 @@ function writeAppFile(filePath, content) {
 }
 
 /**
+ * Verify directory exists in src/app folder
+ */
+function verifyAppFolder(folderPath) {
+  const resolvedFolderPath = path.join(path.resolve(tmp), 'src', 'app', folderPath);
+  return new Promise((resolve, reject) => {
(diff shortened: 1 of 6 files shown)


stirno:
Just realized this was opened under wrong account.. closing/re-opening.

PR #179, opened 2017-06-12
Stop resolving, fix windows guard paths

blackbaud-brandonstirnaman:
Fixes the bathing to use the runtime alias. Also flips `\` to `/` for module loading on windows. Confirmed fix on Win7 and OSX.

### DIFF
--- a/lib/sky-pages-route-generator.js
+++ b/lib/sky-pages-route-generator.js
@@ -141,10 +141,11 @@ function generateDeclarations(routes) {
   return `[\n${declarations}\n]`;
 }
 
-function generateRuntimeImports(routes) {
+function generateRuntimeImports(skyAppConfig, routes) {
+  const alias = skyAppConfig.runtime.spaPathAlias;
   return routes
     .filter(r => r.guard)
-    .map(r => `import { ${r.guard.name} } from '${r.guard.path.replace(/\.ts$/, '')}';`);
+    .map(r =>`import { ${r.guard.name} } from '${alias}/${r.guard.path.replace(/\.ts$/, '')}';`);
 }
 
 function generateProviders(routes) {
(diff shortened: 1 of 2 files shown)


lib/sky-pages-route-generator.js
```diff
@@ -142,10 +142,11 @@ function generateDeclarations(routes) {
   return `[\n${declarations}\n]`;
 }
 
-function generateRuntimeImports(routes) {
+function generateRuntimeImports(skyAppConfig, routes) {
+  const alias = skyAppConfig.runtime.spaPathAlias;
   return routes
     .filter(r => r.guard)
-    .map(r => `import { ${r.guard.name} } from '${r.guard.path.replace(/\.ts$/, '')}';`);
+    .map(r =>`import { ${r.guard.name} } from '${alias}/${r.guard.path.replace(/\.ts$/, '')}';`);
```
Blackbaud-BobbyEarl: Since you were able to make these code changes without having to update any unit tests, seems you've found a gap in our coverage.  :-( 

Would you be comfortable writing a quick unit test to verify spaPathAlias is appended to the importPath?

lib/sky-pages-route-generator.js
```diff
@@ -190,7 +191,7 @@ function extractGuard(file) {
       throw new Error(`As a best practice, only export one guard per file in ${file}`);
     }
 
-    result = { path: path.resolve(file), name: match[1] };
+    result = { path: file.replace(/\\/g, '/'), name: match[1] };
```
blackbaud-brandonstirnaman: The `path.resolve` was basically in there as a stop-gap before that I forgot to take out. We really don't want it there as it'll make the module path absolute.. and would then break deployments too.

Additionally, this path is used for node.js module resolution would show *always* be `/` even on windows.

lib/sky-pages-route-generator.js
blackbaud-brandonstirnaman: on it.

PR #180, opened 2017-06-12
Preload hook gets run before build

Blackbaud-SteveBrush:


Blackbaud-SteveBrush:
@Blackbaud-PaulCrowder: I renamed the module, and adjusted the export to be a module, instead of a single function (`pluginFileProcessor.processFiles()`). Thanks for the feedback.

PR #182, opened 2017-06-13
Stopping same param from getting added more than once

Blackbaud-BobbyEarl:


Blackbaud-BobbyEarl:
Ahh thanks @blackbaud-brandonstirnaman.  I'm a little premature on creating the PR.  I do believe this "solves" https://github.com/blackbaud/skyux2/issues/742, but I'm digging in to the actual root cause at the moment.

Blackbaud-BobbyEarl:
Did some more debugging today with the help of @Blackbaud-PaulCrowder.  This bug presents itself when all of the following are true:

- Consumer is using the `SkyAuthHttp.request` method as opposed to one of the convenience methods, `get`, `post`, etc.
- Consumer is subscribing to the same request multiple times.
- Consumer is not using `Observable.share()` method to make it a hot observable, which means a new request is created for each subscriber.

We decided implementing the logic to not add the same variable to a url via the `getUrl` method is the most consistent fix, but it was important to understand how this bug was happening, which I wasn't entirely sure on when the PR was first submitted.

PR #186, opened 2017-06-16
Plugin File Processor should not check directories

Blackbaud-SteveBrush:


lib/plugin-file-processor.js
```diff
@@ -11,6 +11,10 @@ const processFiles = (skyPagesConfig) => {
   const filePaths = glob.sync(skyPagesConfigUtil.spaPathTempSrc('app', '**', '*.*'));
 
   filePaths.forEach(filePath => {
+    if (fs.lstatSync(filePath).isDirectory()) {
```
Blackbaud-BobbyEarl: Could we instead pass in `nodir` option to `glob.sync`?

PR #187, opened 2017-06-16
Ignore public directory when generating components

Blackbaud-SteveBrush:
Custom component libraries live in the **src/app/public** folder, but this causes problems because Builder automatically declares these components as its own. The component library should instead declare the components in their own custom modules, and import them into the `AppExtrasModule`.

(The unit tests for this felt odd; let me know if there's a better way to do it.)


config/sky-pages/sky-pages.config.js
```diff
@@ -71,6 +71,7 @@ module.exports = {
         },
         command: command,
         componentsPattern: '**/*.component.ts',
+        componentsIgnorePattern: './public/**/*.component.ts',
```
Blackbaud-BobbyEarl: I believe `componentsIgnorePattern` will also have to be added to `runtime/config.ts`.  Also, can we be less specific here and just say `componentsIgnorePattner: './public/'`?

PR #199, opened 2017-06-23
Fixed type error, updated SKY UX

Blackbaud-SteveBrush:


Blackbaud-BobbyEarl:
Just as a point of clarification, we originally defined our return type incorrectly as it's always been `Observable<Response>` from the `Http` `request` method.  https://github.com/angular/angular/blame/3e51a1998304ab6a15e5bea6bc66e7a8c636a8ad/packages/http/src/http.ts#L112

An upgrade to a stricter type-checking caught this.

PR #202, opened 2017-06-27
Fixed `skyux test/watch` performance

Blackbaud-SteveBrush:
- Issue is with `tslint-loader`. (https://github.com/wbuchwalter/tslint-loader/issues/76)
- **Please confirm that `skyux watch --no-coverage` still works as intended.**

Fixes https://github.com/blackbaud/skyux2/issues/834

### UPDATE 1
Setting `typeCheck: false` within `tslint-loader` _does_ make the tests go faster, but it forfeits any of the rules found in **tslint.json** that require type checking (eg, `no-unused-variables`). To see a full list of rules affected, visit the [Rules documentation](https://palantir.github.io/tslint/rules/) and look for rules labeled "Requires Type Info".

The reason it is so slow is because `tslint-loader` is instantiating a fresh compiler object for every single file it checks. The `tslint` CLI does not have this problem, however, because it is only generating a single instance of the TypeScript compiler for all files. While [there is a PR](https://github.com/wbuchwalter/tslint-loader/pull/78) to fix this problem in `tslint-loader` it hasn't been touched since May 22. 

**Options:** 
1. We disable type checking in `tslint-loader` and wait for [tslint-loader#78](https://github.com/wbuchwalter/tslint-loader/pull/78) to be merged, or 
2. Find another method of linting files (such as `ts-loader`), and forfeit `awesome-typescript-loader`.
3. Create a new command `skyux lint` that spawns `./node_modules/.bin/tslint --type-check --project tsconfig.json`, which does the same thing, but would not behave as a webpack loader. We could modify `skyux test` to also call `skyux lint`, so that existing build flows would not be affected.

IMO, Option #3 would have the least impact on existing projects, and would not require that much work to accomplish.

### UPDATE 2
I ended up writing a webpack loader/plugin that handles TSLint for all webpack files. I initially attempted a Karma preprocessor that would lint all files in `src`, but it got a little complex due to the way Karma loops through each file one-by-one. Aside from that, this branch will give us feature parity with the `rc-` branch, will check against all rules in `tslint.json`, and will execute much faster. 

I went down the path of creating a new `skyux lint` command (as you can see with PR #205), but I feel that it would be better suited for a different feature branch, as the requirements for that command are for ALL files in **src**, not just the webpack files. 

Thoughts, @Blackbaud-BobbyEarl?


config/webpack/serve.webpack.config.js
```diff
@@ -121,11 +119,6 @@ function getWebpackConfig(argv, skyPagesConfig) {
     module: {
       rules: [
         {
-          enforce: 'pre',
```
Blackbaud-SteveBrush: Removed this rule because it is called in the common config.

config/webpack/test.webpack.config.js
```diff
@@ -87,7 +76,16 @@ function getWebpackConfig(skyPagesConfig, argv) {
...
           loader: outPath('loader', 'sky-processor', 'preload'),
-          exclude: /node_modules/
+          exclude: excludes
+        },
+        {
+          test: /\.s?css$/,
+          use: ['raw-loader', 'sass-loader']
+        },
+        {
+          test: /\.html$/,
```
Blackbaud-SteveBrush: Updated CSS, SCSS, and HTML rules to look just like common config (no affect on output).

config/webpack/serve.webpack.config.js
Blackbaud-SteveBrush: Removed this rule because it is already listed in "common" config.

config/webpack/test.webpack.config.js
```diff
@@ -108,19 +104,12 @@ function getWebpackConfig(skyPagesConfig, argv) {
           exclude: [/\.e2e\.ts$/]
         },
         {
-          test: /\.css$/,
-          loader: 'raw-loader'
+          test: /\.s?css$/,
+          use: ['raw-loader', 'sass-loader']
```
Blackbaud-SteveBrush: Just some cleanup to match what's in common config.

Blackbaud-BobbyEarl:
@Blackbaud-SteveBrush and @Blackbaud-PaulCrowder... I'm thinking the long-term solution to this problem (including fixing https://github.com/blackbaud/skyux2/issues/411) may take slightly longer than anticipated.  

Any objections on creating a release immediately after either A) accepting this PR or B) creating a new PR that disables type-checking (knowing that's reducing features), but buys us some time to make the right decision.

Blackbaud-SteveBrush:
I'm obviously partial to merging this in as-is, but I will say that I kept the functionality isolated to the **loaders/sky-tslint** folder, so if we need to drop it for something better in the future, it wouldn't be too difficult.

Also, this branch would not have a drop in features for existing users; another plus.

PR #205, opened 2017-06-29
Added `skyux lint` command

Blackbaud-SteveBrush:
**Reason for the change:**

- Type checking (apparently) is only available for TSLint CLI. The "runner" is only executed within the CLI (https://github.com/palantir/tslint/blob/1ec0bace70fa834b33c8851a5768596dc2155373/src/tslint-cli.ts#L24 and https://github.com/palantir/tslint/blob/1ec0bace70fa834b33c8851a5768596dc2155373/src/runner.ts#L160).
- Also, type checking only works when you have a TSLint "project", i.e., a tsconfig.json file, so we can't run a blanket type check on all source files (unless defined in tsconfig.json). https://palantir.github.io/tslint/usage/type-checking/

**What this branch accomplishes:**

- I removed our custom TSLint linter from Webpack entirely, and created a separate spawn to run the type checking through the CLI.
- New command: `skyux lint`. 
- This command is executed during `build`, `e2e`, `test`, and `watch`. It is ignored for `serve`.
- I had to make a few syntactic changes to some of the builder's source code because of the new linting process.

Addresses: https://github.com/blackbaud/skyux2/issues/887


tsconfig.json
```diff
@@ -18,8 +19,19 @@
       "node",
       "jasmine",
       "fontfaceobserver"
-    ]
+    ],
+    "baseUrl": ".",
+    "paths": {
+      "@blackbaud/skyux-builder/*": [
+        "./node_modules/@blackbaud/skyux-builder/*",
```
Blackbaud-SteveBrush: Added this to tell TS about the webpack aliases.
https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping

tsconfig.json
```diff
@@ -18,8 +19,19 @@
...
+        "./*"
+      ]
+    }
   },
+  "exclude": [
+    "node_modules",
+    "**/*.aot.ts"
```
Blackbaud-SteveBrush: Ignore TS's `Cannot find module './ngfactory/app/app.module.ngfactory'.` within the **main-internal.aot.ts** file.

Blackbaud-SteveBrush:
@Blackbaud-PaulCrowder This one is ready for another look.

PR #216, opened 2017-07-07
Moved Angular to `devDependencies`

Blackbaud-SteveBrush:
Removing Angular as a dependency so that SKY UX can determine the correct version of Angular to use. Hopefully this will address the issue of builds not working due to multiple versions of Angular existing in `node_modules`.


Blackbaud-SteveBrush:
Closing, in favor of https://github.com/blackbaud/skyux2/pull/876

PR #219, opened 2017-07-13
Serve and open built files

Blackbaud-BobbyEarl:
This modification exposes the `-l` or `--launch` flag to `skyux build`, which will accept `local` and `host` as values.  Essentially, it easily allows a "built" SPA to be served.

https://github.com/blackbaud/skyux2/issues/562
https://github.com/blackbaud/skyux2/issues/890


cli/utils/stage-library-ts.js
```diff
@@ -66,10 +66,10 @@ function getFileContents(filePath) {
   switch (path.extname(filePath)) {
     case '.scss':
       contents = compileSass(filePath);
-      break;
+    break;
```
Blackbaud-BobbyEarl: It's unclear how this got in to start with, but the linter complained to be about the spacing.

Blackbaud-BobbyEarl:
Hey @Blackbaud-SteveBrush and @Blackbaud-PaulCrowder, this one is ready for review.

The factoring out of `server.js` and `browser.js` so they could be shared across different CLI commands, unfortunately meant I had to move quite a bit of the hoisting I would doing for tests to different locations (which is why the PR looks so large).

Let me know if you have any questions or want to talk through any of it.

Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl: Launching on localhost works great!

```
skyux build --launch local
skyux build -l local
```

However, when using `--launch host`, it seems that locales aren't being parsed correctly:

screen shot 2017-08-18 at 2 27 32 pm

Blackbaud-SteveBrush:
One last comment, and this is just a matter of preference: it might be nice to have a default value, so that consumers could type `skyux build -l`.

Blackbaud-BobbyEarl:
Thanks for the feedback @Blackbaud-SteveBrush.  Based on it, I've implemented the following changes:

- Adjusted the flag to `--serve` or `-s`.  This frees up the `--launch` flag to be used exactly as it is with `skyux serve`, accepting `none`, `local`, and `host` (default) as values.  This is also handy when running the command more than once, meaning you can `skyux build -s` the first time, and `skyux build -s -l none` on subsequent runs.
- Fixed the resources/assets not loading when launching `host`.

These together mean you can run some pretty neat commands, for example to launch the local url in safari... `skyux build -s -l local -b safari`.  Anyway, it's ready for review again.

cli/e2e.js
```diff
@@ -91,6 +85,7 @@ function spawnProtractor(chunks, port, skyPagesConfig) {
     protractorPath,
     [
       getProtractorConfigPath(),
+      `--disableChecks`,
```
Blackbaud-BobbyEarl: FYI, adding this check to solve the "unknown arguments" warning (soon to be error) that's shown.  https://github.com/angular/protractor/blob/master/CHANGELOG.md#features-3

cli/build.js
```diff
@@ -114,45 +118,65 @@ function cleanupAot() {
...
+  const isAot = skyPagesConfig &&
+    skyPagesConfig.skyux &&
+    skyPagesConfig.skyux.compileMode === 'aot';
+
+  if (lintResult.exitCode > 0) {
+    process.exit(lintResult.exitCode);
+  } else if (argv.serve) {
+    buildServe(argv, skyPagesConfig, webpack, isAot);
+  } else {
+    return buildPromise(argv, skyPagesConfig, webpack, isAot);
```
Blackbaud-SteveBrush: Curious why we're returning the promise for `buildPromise`, but not for `buildServe`? Would it be appropriate to also catch any server errors?

```
function buildServe(argv, skyPagesConfig, webpack, isAot) {
  return server.start(skyPagesConfigUtil.getAppBase(skyPagesConfig))
    .then(port => {
      argv.assets = argv.assets || `https://localhost:${port}`;
      return buildPromise(argv, skyPagesConfig, webpack, isAot);
    })
    .then(stats => browser(argv, skyPagesConfig, stats, port));
}
```

cli/e2e.js
```diff
@@ -218,7 +177,7 @@ function e2e(argv, skyPagesConfig, webpack) {
     return killServers(0);
   }
 
-  spawnServer()
+  server.start()
```
Blackbaud-SteveBrush: So much cleaner; love it.

config/protractor/protractor-dev.conf.js
```diff
@@ -19,36 +20,44 @@ let config = {
...
+        resolve();
 
-    return new Promise((resolve, reject) => {
-      common.rimrafPromise(common.tmp)
-        .then(resolve)
-        .catch(reject);
+      } else {
+
+        const url = 'https://github.com/blackbaud/skyux-template';
+        const branch = 'master';
```
Blackbaud-SteveBrush: I pointed the branch to `builder-dev` (a protected branch) so that we could adjust the template as needed without affecting what consumers are cloning when typing `skyux new`. Unsure how you feel about this.

Blackbaud-SteveBrush: https://github.com/blackbaud/skyux-template/tree/builder-dev

Blackbaud-SteveBrush: (This is also helpful for when we have various versions of SKY UX that need to be tested during a Travis build.)

Blackbaud-BobbyEarl: Changing this wasn't intentional... chalking it up to a bad merge.

Learning about it though, I'm torn.  It seems nice for development for new features in builder that may require a change to the template.  On the other hand, running builder's e2e tests is used to test that `skyux new` runs as intended for consumers, in which case they'd be using the `master` branch.

I'll change it back to `builder-dev` for the sake of this PR and we can talk more about it later.

cli/build.js
Blackbaud-BobbyEarl: If `skyux build` fails outside of linting, [the ProcessExitCode plugin](https://github.com/blackbaud/skyux-builder/blob/ecce9dba359157f7e7521b05eaf3135d491d189f/config/webpack/common.webpack.config.js#L145) calls `process.exit`.

I was really only concerned with returning a promise from `buildPromise` knowing that `cli/e2e.js` imports `build`, which wouldn't be passing the `serve` command.  It definitely makes sense to make the signature the same (always returning a promise) regardless of the arguments.

**tl;dr** I'm going to push an update to always have `build` return a promise, but `process.exit` is handled in the webpack plugin.

Blackbaud-SteveBrush: Gotcha. That makes sense!

Blackbaud-BobbyEarl:
Any feedback before this gets merged in @Blackbaud-PaulCrowder?

PR #220, opened 2017-07-13
Naming convention for "utility" directories

Blackbaud-SteveBrush:
To my knowledge, we have three "utility" directories in a given SPA:
- `src/app/lib`
- `src/app/public`
- `src/app/shared`

Each of these directories serves a specific purpose and is handled in a special way by Builder. This pull request is more to facilitate a discussion around convention, and if there's room to _visually_ separate the folder names from the standard "routing" directories. (Much like the SASS convention, where files that are not entry points are prefixed with an underscore, `_variables.scss`.)

**For example, a SPA's source might look like this, today:**

screen shot 2017-07-13 at 11 08 01 am

**Ideally, there would be some visual indicator that the utility directories have a special purpose:**

screen shot 2017-07-13 at 11 08 35 am


Blackbaud-BobbyEarl:
Using an underscore's not going to work as that's what we use to delineate folders that are route params.  https://github.com/blackbaud/skyux-builder/blob/master/lib/sky-pages-route-generator.js#L74

Blackbaud-SteveBrush:
Hm, doesn't appear to be a "safe" alternative, at least for a leading character: https://superuser.com/questions/358855/what-characters-are-safe-in-cross-platform-file-names-for-linux-windows-and-os

PR #225, opened 2017-07-14
Removed preloader from AoT Webpack config

Blackbaud-SteveBrush:
Since the preloader is executed on the file system during an AoT build, the loader is run twice: once, during the [webpack build](https://github.com/blackbaud/skyux-builder/blob/master/config/webpack/common.webpack.config.js#L95), and second, in the [`stageAot`](https://github.com/blackbaud/skyux-builder/blob/master/cli/build.js#L108) method.

This duplicate behavior causes plugins to process files twice during AoT, causing things like this (the double-escaped brackets):
https://developer.blackbaud.com/skyux2/learn/reference/configuration

screen shot 2017-07-13 at 11 28 02 pm

### DIFF
--- a/config/webpack/build-aot.webpack.config.js
+++ b/config/webpack/build-aot.webpack.config.js
@@ -19,6 +19,14 @@ function getWebpackConfig(skyPagesConfig) {
   let commonConfig = common.getWebpackConfig(skyPagesConfig);
   commonConfig.entry = null;
 
+  // Since the preloader is executed against the file system during an AoT build,
+  // we need to remove it from the webpack config, otherwise it will get executed twice.
+  commonConfig.module.rules = commonConfig.module.rules
+    .filter((rule) => {
+      const isPreloader = /\/sky-processor\//.test(rule.loader);
+      return (!isPreloader);
+    });
+
   return webpackMerge(commonConfig, {
     entry: {
       polyfills: [skyPagesConfigUtil.spaPathTempSrc('polyfills.ts')],
--- a/test/config-webpack-build-aot.spec.js
+++ b/test/config-webpack-build-aot.spec.js
@@ -193,4 +193,40 @@ describe('config webpack build-aot', () => {
     });
   });
 
+  it('should remove the sky-processor loader from the rules array', () => {
+    const f = './common.webpack.config';
+    mock(f, {
+      getWebpackConfig: () => ({
+        module: {
+          rules: [
+            {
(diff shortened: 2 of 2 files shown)


config/webpack/build-aot.webpack.config.js
```diff
@@ -19,6 +19,14 @@ function getWebpackConfig(skyPagesConfig) {
   let commonConfig = common.getWebpackConfig(skyPagesConfig);
   commonConfig.entry = null;
 
+  // Since the preloader is executed against the file system during an AoT build,
+  // we need to remove it from the webpack config, otherwise it will get executed twice.
+  commonConfig.module.rules = commonConfig.module.rules
```
Blackbaud-BobbyEarl: I like the comment.  Thanks for that.  :-)

Any thoughts on shortening the syntax?  Obviously doing the exact same thing, just removed some extraneous parenthesis and removed the `const`.  I'm pretty impartial on this, just thought the syntax looked a little weird.

```
 commonConfig.module.rules = commonConfig.module.rules
  .filter(rule => !/\/sky-processor\//.test(rule.loader));
```

test/config-webpack-build-aot.spec.js
```diff
@@ -193,4 +193,40 @@ describe('config webpack build-aot', () => {
...
+
+    config.module.rules.forEach((rule) => {
+      if (found) {
+        return;
+      }
+
+      found = /\/sky-processor\//.test(rule.loader);
+    });
+
+    expect(found).toEqual(false);
```
Blackbaud-BobbyEarl: Missing `mock.stop(f);`.

config/webpack/build-aot.webpack.config.js
Blackbaud-SteveBrush: LOL I had it in this format earlier, but thought setting the boolean made it easier to read.

test/config-webpack-build-aot.spec.js
Blackbaud-SteveBrush: Good catch. Added `mock.stopAll();` to the `afterEach()`.

config/webpack/build-aot.webpack.config.js
colbywhite: I'd say using the boolean _does_ make it easier to read. You could make it even easier to read by defining the function into a well-named variable. Simply my two outside-looking-in cents.

PR #232, opened 2017-07-18
Added a check for a route fragment in the navigationEnd event 

Blackbaud-BrandonJones:
The navigationEnd subscription was causing pages to scroll to the top of the page on every navigation event.  This was breaking in-page navigation to fragments on the page.   With this change, it allows apps to scroll to elements on their page based on the fragment, as is expected from in page anchor links.


src/app/app.component.spec.ts
```diff
@@ -54,7 +54,12 @@ describe('AppComponent', () => {
           events: {
             subscribe: handler => subscribeHandler = handler
           },
-          navigateByUrl: url => navigateByUrlParams = url
+          navigateByUrl: url => navigateByUrlParams = url,
+          parseUrl: url =>   {
```
Blackbaud-SteveBrush: Check your spaces count after the lambda, and after the `undefined`, below.

Blackbaud-SteveBrush:
@Blackbaud-PaulCrowder I don't see anything wrong with this one. You?

Blackbaud-PaulCrowder:
Seems fine to me.

PR #234, opened 2017-07-18
Set the assets URL during e2e tests

Blackbaud-PaulCrowder:


Blackbaud-SteveBrush:
@Blackbaud-PaulCrowder The feature works great; tested all CLI commands on various SPAs.

I didn't see a unit test where we're passing in a relative value to `getAssetsUrl()`. Is that important?

Something like...

**lib-assets-processor.spec.js**
```
const url = processor.getAssetsUrl(
    {
      runtime: {
        app: {
          base: 'test'
        }
      }
    },
    'https://example.com/',
    '../' // <-- passing in a relative value...
  );
```

PR #237, opened 2017-07-21
Updated font families in style loader

Blackbaud-SteveBrush:
Since the font families have changed in SKY UX rc.7, the style loader doesn't resolve in time and causes Jasmine tests to timeout.


Blackbaud-ScottFreeman:
Ship it :)

PR #239, opened 2017-07-21
Added templating ability to test resources service

Blackbaud-TrevorBurch:
blackbaud/skyux2#925


Blackbaud-PaulCrowder:
The `formatText` method is copied and pasted from the non-test version of this class.  I'd like to see that factored out into another class that can be consumed by both.  We can either accept this as-is and refactor it after it's merged or @Blackbaud-TrevorBurch can take a stab at doing the refactoring.  I'd just create a `SkyAppFormat` for now and put the function in that class.  Maybe later we can move the logic to the SKY UX 2 component library since we have an equivalent service in SKY UX 1.

Blackbaud-TrevorBurch:
@Blackbaud-PaulCrowder where would you want that class to live? Just inside `runtime`?

Blackbaud-TrevorBurch:
@Blackbaud-PaulCrowder @Blackbaud-SteveBrush I took a stab at the suggested refactor. Let me know if you need anything else.

Blackbaud-TrevorBurch:
Digging into these errors.

Blackbaud-PaulCrowder:
One thing I would suggest is to not export `SkyAppFormat` in the runtime index.ts barrel.  Let's keep this an internal implementation detail for now and later we can evaluate whether we make it part of the SKY UX core component library or even an external NPM package that can be consumed by both the component library and builder.

Blackbaud-TrevorBurch:
My novice status here is getting me :) This appears to be because of me not yet figuring out where to provide the SkyAppFormat so that it can be injected correctly. Still working through that

Blackbaud-TrevorBurch:
Here goes nothing :)

Blackbaud-SteveBrush:
If @Blackbaud-PaulCrowder approves this, I'm fine with it! 

However, it'd be my preference if we came up with a more descriptive name for either the service or the service's method. I had to look at the source code to figure out what it was doing (binding variables to a string template).

All-in-all, this works great; using the term "template" or "binding" somewhere in the service/method name would make it even better. I'm willing to concede my opinion, however, if this is indeed a common usage of the word "format".

Blackbaud-TrevorBurch:
@Blackbaud-SteveBrush Looking at the Sky UX 1 stuff it looks like the method name and class name follow what was done there. I'll leave the decision up to @Blackbaud-PaulCrowder but I like the consistency currently here so that it matches Sky 1 if this class ends up being pulled out and put into Sky UX 2.

Blackbaud-PaulCrowder:
I called it "format" in SKY UX 1 because of its similarities to `String.Format()` in .NET and `String.format()` in Java.  Both of those functions take a template string as its first argument and replaces the tokens in that string with the remaining arguments based on the order of those arguments.

PR #245, opened 2017-07-25
Added ability to specify a permission scope to SkyAuthHttp

Blackbaud-PaulCrowder:


Blackbaud-SteveBrush:
I'm assuming if multiple scopes were needed, we could pass in a comma-separated string, `.withScope('id,first_name,last_name')`, for example?

Blackbaud-PaulCrowder:
When SKY UX worked with the permissions and auth teams, we decided we would not support accepting multiple permission scopes when retrieving a token.  If a caller needs more permissions than an existing permission scope already contains, they will need to register a new permission scope with the permissions service.

runtime/auth-http.ts
```diff
@@ -16,26 +16,73 @@ import { Observable } from 'rxjs/Observable';
...
+   * @param permissionScope The permission scope to include.
+   */
+  public withScope(permissionScope: string): SkyAuthHttp {
+    // There was no clean way to allow permission scope to be specified with all the various
+    // convenience methods such as get(), post(), etc. that the base Http class provides,
+    // so this chainable method just creates a new instance of SkyAuthHttp with a permissionScope
+    // property set.  When chained, the end call would look something like this:
+    // http.withScope('abc').get(url);
+
+    const http = new SkyAuthHttp(
```
Blackbaud-SteveBrush: I like this; prevents the scope from being applied to all subsequent requests.

PR #256, opened 2017-08-15
Added ability to bypass e2e if no files or directory exist.

blackbaud-brandonhare:


Blackbaud-BobbyEarl:
Thanks for the PR @blackbaud-brandonhare.  I was thinking something like the following:

```
const glob = require('glob');

...

function e2e() {
  const specsPath = path.resolve(process.cwd(), 'e2e/**/*.spec.ts');
  const specsGlob = glob.sync(specsPath);

  if (specsGlob.length === 0) {
    logger.info(`No spec files located.  Stopping command from running.`);
    return killServers(0);
  }
 
  [method definition that currently exists...]
}
```

cli/e2e.js
```diff
@@ -206,9 +207,17 @@ function spawnBuild(argv, skyPagesConfig, webpack) {
  * @name e2e
  */
 function e2e(argv, skyPagesConfig, webpack) {
+  const specsPath = path.resolve(process.cwd(), 'e2e/**/*.spec.ts */');
```
Blackbaud-BobbyEarl: The glob pattern here has an extra trailing ` */`.  I forgot the `e2e-` to the pattern in my example.

The final glob pattern should be `e2e/**/*.e2e-spec.ts`

cli/e2e.js
```diff
@@ -206,9 +207,17 @@ function spawnBuild(argv, skyPagesConfig, webpack) {
...
+  const specsGlob = glob.sync(specsPath);
```
Blackbaud-BobbyEarl: I know I had them in this order in my example, but looking at it now, I think the two `const` definitions would be better after creating start and adding the listener.  So...

```
function e2e(argv, skyPagesConfig, webpack) {
  start = new Date().getTime();
  process.on('SIGINT', killServers);

  const specsPath = path.resolve(process.cwd(), 'e2e/**/*.e2e-spec.ts');
  const specsGlob = glob.sync(specsPath);

  if (specsGlob.length === 0) {
    logger.info('No spec files located. Stopping command from running.');
    return killServers(0);
  }

  ...
```

PR #257, opened 2017-08-16
Specifying host when listening for WebpackDevServer

Blackbaud-BobbyEarl:
Fixes https://github.com/blackbaud/skyux2/issues/717


Blackbaud-BobbyEarl:
Windows was the only affected system.  Ultimately, I believe https://github.com/indexzero/node-portfinder/issues/39 is the problem, but explicitly passing `localhost` into `webpack-dev-server` resolves it (they use `portfinder` under the hood too).

PR #265, opened 2017-08-31
Using backticks to render filename

Blackbaud-BobbyEarl:


config/sky-pages/sky-pages.config.js
```diff
@@ -51,7 +51,7 @@ module.exports = {
         filePath: this.spaPath(`skyuxconfig.json`)
       },
       {
-        fileName: 'SPA skyuxconfig.${command}.json',
+        fileName: `SPA skyuxconfig.${command}.json`,
```
Blackbaud-SteveBrush: Was there not a unit test for this?

Blackbaud-BobbyEarl: There was a test to validate the order of merging, but not that the logging was correct.  I'll update the text to validate the log as well.

PR #270, opened 2017-09-08
Feature help library

Blackbaud-BrandonJones:
Working demo with setup instructions can be found here:
https://github.com/Blackbaud-BrandonJones/bb-help-test-skyux2


Blackbaud-BobbyEarl:
Is the description of this PR still accurate @Blackbaud-BrandonJones?

Blackbaud-BrandonJones:
Nope sorry, will update.  This PR is now entirely self sufficient.

Blackbaud-BrandonJones:
Updated and ready for your review @Blackbaud-BobbyEarl

runtime/config.ts
```diff
@@ -47,7 +66,7 @@ export interface SkyuxConfig {
   cssPath?: string;
   command?: string;
   compileMode?: string;
-  help?: any;
+  help?: HelpWidgetConfig;
```
bobbyearl: I'm curious about the creation of this interface.  We'd previous used interfaces for the auth and help clients.  We decided against that; however, as somewhat painfully tied those libraries to builder, but without any benefit (consumers never use the interface).

src/app/app.component.ts
```diff
@@ -74,6 +74,7 @@ export class AppComponent implements OnInit {
     private windowRef: SkyAppWindowRef,
     private config: SkyAppConfig,
     private styleLoader: SkyAppStyleLoader,
+    private helpInitService: HelpInitializationService,
```
bobbyearl: Could we use a similar approach to the optional `searchProvider` below this line, and mark the service as optional + check for it's existence for using it? I believe this would allow us to conditionally inject it in `lib/sky-pages-module-generator.js`, thereby not including the library for SPAs that don't have `help` defined in `skyuxconfig.json`.  Thoughts?

Blackbaud-BrandonJones: Good thoughts! I'll see what I can do.

runtime/config.ts
Blackbaud-BrandonJones: That makes sense.  I added it because I was using it for the Help Library and thought builder may want to also make use of it.  That being said, having to update it in multiple places should one change would be a pain, I am fine with removing it.

lib/sky-pages-module-generator.js
```diff
@@ -83,6 +103,12 @@ function getSource(skyAppConfig) {
     'AppExtrasModule'
   ];
 
+  if (skyAppConfig.skyux.help) {
+    nodeModulesImports.push(`import { BBHelpModule } from '@blackbaud/skyux-lib-help';`);
```
Blackbaud-BobbyEarl: I'm nitpicking, but can you name this `nodeModuleImports` so the case matches `runtimeModuleImports`?

PR #275, opened 2017-09-15
Updated linter line length to something less draconian.

blackbaud-brandonhare:


tslint.json
```diff
@@ -82,7 +82,7 @@
     ],
     "max-line-length": [
       true,
-      100
+      200
```
Blackbaud-SteveBrush: I'm going to disagree here, but will defer to @Blackbaud-BobbyEarl. The line length might seem too short on a widescreen monitor, but shorter lines are easier to read and enforce fewer tab structures.

Obviously this is opinion based, but I think the consensus is 80-100 characters.

Interesting read: https://stackoverflow.com/questions/110928/is-there-a-valid-reason-for-enforcing-a-maximum-width-of-80-characters-in-a-code

blackbaud-brandonhare: I understand the rationale, but for me personally, it is really obnoxious with strings. I don't think I should have to modify my string literals to be concatinated or array.joined to pass a linter rule. There has to be some kind of happy medium between the two.

blackbaud-brandonhare: I could also pepper my code with `//Ignore rule`, but that seems just as bad.

Blackbaud-SteveBrush:
```
const str = [
  'a',
  'long',
  'string'
].join('');
```
^ To me, that's a thing of beauty :)

blackbaud-brandonstirnaman: I'm a huge proponent of longer lines than we allow today. I generally fit 120-140 characters per line without scrolling on my laptop.

Blackbaud-LuisBello:
i endorse this pr :us:

PR #277, opened 2017-09-21
Having runCommand return a boolean

Blackbaud-BobbyEarl:
Via https://github.com/blackbaud/skyux-cli/pull/131, `skyux-cli` will become more informed regarding if a module "handled" the command passed via `runCommand`.  This PR satisfies that requirement.


test/index.spec.js
```diff
@@ -57,15 +57,18 @@ describe('@blackbaud/skyux-builder', () => {
     });
   });
 
-  it('should handle unknown command', () => {
+  it('should return false unknown command', () => {
```
Blackbaud-PaulCrowder: The name of this test should probably be "should return false _for_ unknown command".

Blackbaud-BobbyEarl: "for" sure.  :-)  Updated.

Blackbaud-SteveBrush: Forget the "for"? :)

PR #278, opened 2017-09-24
Added config flag to test and e2e

Blackbaud-BobbyEarl:


Blackbaud-SteveBrush:
Adding this comment, so we don't forget to address it: "If the SPA does not have an `e2e` folder, the custom config will never run (even though it might have `*.visual-spec.ts` files)."

Blackbaud-BobbyEarl:
After talking to @Blackbaud-SteveBrush and @Blackbaud-PaulCrowder, we've decided to be a little more explicit in how we allow configuration overrides.  Ultimately this allows to accomplish the same behavior, but in a more first-class way as well as not having karma be a leaky abstraction.

PR #288, opened 2017-10-02
Updated merge library, added tests for overrides and recursive merging.

blackbaud-brandonhare:
Fixes https://github.com/blackbaud/skyux2/issues/949

PR #292, opened 2017-10-05
[HOLD] Using lodash.merge for all merges (except webpack).

Blackbaud-BobbyEarl:
Fixes https://github.com/blackbaud/skyux2/issues/1148

### DIFF
--- a/cli/build.js
+++ b/cli/build.js
@@ -2,7 +2,7 @@
 'use strict';
 
 const fs = require('fs-extra');
-const merge = require('lodash.merge');
+const merge = require('../utils/merge');
 
 const skyPagesConfigUtil = require('../config/sky-pages/sky-pages.config');
 const generator = require('../lib/sky-pages-module-generator');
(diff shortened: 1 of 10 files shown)


Blackbaud-BobbyEarl:
Also thanks to @blackbaud-brandonhare for tracking down the original bug in #288.

For future clarity, other alternatives and their pitfalls explained below:

1.  https://www.npmjs.com/package/merge For reasons that aren't entirely clear, this merge library was unable to have the override merge delete a key.  So if default was `{ test: { inner: true } }` was merged override `{ test: false }`, `test` was still set.  This behavior was documented in the original bug.

2.  https://www.npmjs.com/package/webpack-merge This almost worked, except for the `plugin` property in `skyuxconfig.json`.  By design, those arrays are concatenated in order to work with Webpack.

3.  https://www.npmjs.com/package/lodash.merge This was the best choice as we could pull in just the merge module, instead of all of lodash.  It correctly handled recursively merging as well as deleting keys.

Blackbaud-SteveBrush:
Related? https://github.com/blackbaud/skyux2/issues/1181

PR #294, opened 2017-10-09
Using opacity instead of visibility for FOUC

Blackbaud-SteveBrush:
Addresses: https://github.com/blackbaud/skyux2/issues/902

[Opacity allows for events to be handled](https://stackoverflow.com/questions/272360/does-opacity0-have-exactly-the-same-effect-as-visibilityhidden), which is what the issue above is asking for.

The alternative would be to require that the consuming SPA import the `SkyStyleLoader` from runtime and handle the event after the promise resolves.


Blackbaud-BobbyEarl:
Hey @Blackbaud-SteveBrush and @Blackbaud-MichaelSpires.  This has been merged, but I'll soon be accepting a PR from @Blackbaud-PaulCrowder to undo it.  In it's place, he's implemented a service with an event you can subscribe to.

We found that while events are able to be handled, it also means you can interact with content on the screen that's not visible.  This could lead some unintended behavior.

PR #295, opened 2017-10-09
Hotfix merge arrays

blackbaud-brandonhare:


cli/build.js
```diff
@@ -2,7 +2,7 @@
 'use strict';
 
 const fs = require('fs-extra');
-const merge = require('lodash.merge');
+const merge = require('../utils/merge-utils');
```
Blackbaud-SteveBrush: If we go this route, we can probably forego the added `-utils` suffix on the file, since it's already in the utils folder.

PR #298, opened 2017-10-11
Two fixes for our spa's integration with the GoodData library.

Blackbaud-MikitaYankouski:
Rxjs library update to match the version of GoodData and adding a require to reflect-metadata so our unit tests are able to run and produce code coverage.


package.json
```diff
@@ -84,7 +84,7 @@
     "raw-loader": "0.5.1",
     "reflect-metadata": "0.1.10",
     "remap-istanbul": "0.9.5",
-    "rxjs": "5.4.2",
```
Blackbaud-BobbyEarl: Hey @Blackbaud-MikitaYankouski, in your testing, did you run into any problems upgrading to this version of rxjs.  Our dependency chain isn't crystal clear, and we've ran into problems when upgrading the dependency in builder, but not upgrading it in the skyux2 components.

https://github.com/blackbaud/skyux2/blob/master/package.json#L108

Blackbaud-MikitaYankouski: Just re-ran the all the tests and haven't ran into any issues!

utils/spec-bundle.js
```diff
@@ -30,6 +30,8 @@ require('zone.js/dist/sync-test');
 require('zone.js/dist/proxy');
 require('zone.js/dist/jasmine-patch');
 
+require('reflect-metadata');
```
Blackbaud-SteveBrush: @Blackbaud-MikitaYankouski Mind providing more information on how this affects your unit tests?

package.json
Blackbaud-SteveBrush: The template also works with these changes. A point of note: UX does not directly depend on rxjs, so the mixmatched versions shouldn't affect a SPA.

utils/spec-bundle.js
Blackbaud-MikitaYankouski: Without this line the unit tests do not run and the code coverage results aren't printed so our build ends up failing as well.

package.json
Blackbaud-MikitaYankouski: We are using a third party library that is required by our app that is more heavily dependent on rxjs, and they use this higher version and having two version in our node_modules gave us build errors.

package.json
```diff
@@ -112,4 +112,4 @@
     "proxyquire": "1.8.0",
     "rimraf": "2.6.1"
   }
-}
+}
```
Blackbaud-PaulCrowder: I noticed your editor removed the trailing newline character from both these files.  Those newlines need to be preserved for TSLint and so would cause problems if you had edited a .ts file.  The project's .editorconfig file specifies that a newline should be added to a file on save, which makes me think your editor is not configured to recognize the .editorconfig file.  If you're using VSCode, you can install the EditorConfig plugin to fix this.

PR #306, opened 2017-10-19
Wrapped the first and last parts with path separators so that only distinct parts are matched

Blackbaud-PaulCrowder:
Fixes https://github.com/blackbaud/skyux2/issues/1222

### DIFF
--- a/test/config-webpack-test.spec.js
+++ b/test/config-webpack-test.spec.js
@@ -56,25 +56,33 @@ describe('config webpack test', () => {
   });
 
   it('should match on whole folder names when excluding folders from code coverage', () => {
+    function createTestPaths(items) {
+      return [
+        ...items,
+        // Add backslash variants of all provided paths.
+        ...items.map(item => item.replace(/\//g, '\\'))
+      ];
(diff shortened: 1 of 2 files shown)


test/config-webpack-test.spec.js
```diff
@@ -17,37 +15,85 @@ describe('config webpack test', () => {
...
+      '/src/app/some-other-index.ts'
+    ];
+
+    const disallowedPaths = [
+      '/home/node_modules/some-folder/test.ts',
+      '/home/testing/src/test.ts',
+      '/src/app/fixtures/test.ts',
+      '/src/app/testing/test.ts',
+      '/src/app/lib/test.ts',
+      '/src/app/index.ts'
```
Blackbaud-SteveBrush: Should we also test Window's backslashes? 
```
const disallowedPaths = [
  ...
  String.raw`\src\app\lib\test.ts`
];
```

Blackbaud-PaulCrowder: Yeah, I'll add some.

PR #309, opened 2017-10-25
[HOLD] Removed flag for SkyAuthHttp injection.

blackbaud-brandonhare:
I made this PR to spark some discussion. Steve and I have discussed this already, but Bobby, I wanted to get your input. I'll grab you once you get in.

### DIFF
--- a/lib/sky-pages-module-generator.js
+++ b/lib/sky-pages-module-generator.js
@@ -42,6 +42,7 @@ function getSource(skyAppConfig) {
     'SkyAppConfig',
     'SkyAppRuntimeConfigParams',
     'SkyAppWindowRef',
+    'SkyAuthHttp',
     'SkyAuthTokenProvider',
     'SkyAppLinkDirective',
     'SkyAppStyleLoader',
@@ -52,6 +53,11 @@ function getSource(skyAppConfig) {
     'SkyAppWindowRef',
     'SkyAppStyleLoader',
     'SkyAuthTokenProvider',
+    `{
+      provide: SkyAuthHttp,
+      useClass: SkyAuthHttp,
+      deps: [XHRBackend, RequestOptions, SkyAuthTokenProvider, SkyAppConfig]
+    }`,
     `{
       provide: SkyAppConfig,
       deps: [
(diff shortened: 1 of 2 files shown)


Blackbaud-BobbyEarl:
In order to keep things documented can you add a little context around this change? What’s the scenario where you’re always needing `SkyAuthHttp` provided?

blackbaud-brandonhare:
For my specific use case, I'm using it in my SearchResultsProvider to make requests to my Stache Search Query endpoint. That query conditionally requires the request to be authenticated depending on the status of the site (internal or public). The problem is, if Auth is set to false, the SkyAuthHttp provider doesn't get injected, and it will crash any Stache site that uses the new provider, which will be all of them. I'm not certain this is the ideal approach, but I wanted to create the PR just so we could discuss it further.

blackbaud-brandonhare:
Hmm. I actually found a better solution in my provider. Closing this.

Blackbaud-BobbyEarl:
The other solution not work out @blackbaud-brandonhare?

blackbaud-brandonhare:
It does @Blackbaud-BobbyEarl, but @Blackbaud-SteveBrush wanted me to reopen it for discussion about the SkyAuthHttp and Http in general. I can close it if we want to take that elsewhere.

Blackbaud-BobbyEarl:
Nah fine by me to leave it open.  Let's discuss away.  :-)

Blackbaud-SteveBrush:
Related: https://github.com/blackbaud/skyux2/issues/1252

If the library wishes to use `SkyAuthHttp` (and even provide it statically through their own service), it won't work because the auth http service is never provided (unless `auth=true` in **skyuxconfig.json**). This essentially makes `SkyAuthHttp` unusable in component libraries.

@Blackbaud-BobbyEarl: I vote to merge this in, if possible, to allow libraries to use `SkyAuthHttp` in their implementations regardless of the `auth` setting.

Blackbaud-SteveBrush:
@blackbaud-brandonhare Something I just thought about: curious if this will affect the auto-redirect to the login endpoint?

blackbaud-brandonhare:
@Blackbaud-SteveBrush Hmm. That's a good question. It's worth investigating that.

blackbaud-brandonhare:
@Blackbaud-SteveBrush I just ran `e2e` and everything worked as expected, so I think we're good Steve.

Blackbaud-SteveBrush:
Let's close this until we've figured out how we want libraries to consume the `runtime` directory.

PR #311, opened 2017-10-26
Help E2E integration test

Blackbaud-BrandonJones:
Integration test for the help widget being used with a full page modal.

Adjustments to the `common.js` methods:
new methods for adding and reverting the `app-extras.module` file, so you can create a custom module for e2e tests.
Updated the `removeFolder` method to be recursive and remove all folder contents before trying to remove the folder itself.  This was causing a promise rejection if the folder was empty.


e2e/shared/common.js
```diff
@@ -248,10 +255,35 @@ function removeAppFolder(folderPath) {
   const resolvedFolderPath = path.join(path.resolve(tmp), 'src', 'app', folderPath);
   return new Promise((resolve, reject) => {
     if (fs.existsSync(resolvedFolderPath)) {
-      fs.rmdirSync(resolvedFolderPath);
+      /**
```
Blackbaud-BobbyEarl: Have you thought about importing `fs-extra` and calling `emptyDirSync` instead?

Blackbaud-BrandonJones: I did not know that was a thing.   But I suppose that would be less code.

Blackbaud-BrandonJones: Good call on using `fs-extra` saved a ton of code and dried up some of the methods we had going.

e2e/shared/common.js
```diff
@@ -14,9 +14,11 @@ const tmp = './.e2e-tmp/';
 const cwdOpts = { cwd: tmp };
 
 const skyuxConfigPath = path.resolve(process.cwd(), tmp, 'skyuxconfig.json');
+const appExtrasPath = path.resolve(process.cwd(), tmp, 'src/app', 'app-extras.module.ts');
```
Blackbaud-BobbyEarl: Non-blocking comment, but the last two params could be combined into a single string.

Blackbaud-BrandonJones:
@Blackbaud-BobbyEarl updated this to reflect your non blocking change, also updated with master.

PR #330, opened 2017-11-13
Added ability to generate a component from the CLI

Blackbaud-PaulCrowder:


Blackbaud-PaulCrowder:
https://github.com/blackbaud/skyux2/issues/419

Blackbaud-BobbyEarl:
LGTM. General discussion which I wouldn't necessarily consider non-blocking feedback.  

Is there an opportunity to combine this functionality with `skyux new`.  Perhaps either by running this command as part of the `skyux new` process, so that a new site will always contain the latest example of a component.

Or perhaps flip that, and keep the logic here but have it pull the source files from the template?

Blackbaud-PaulCrowder:
I think it makes sense to try to keep the generated components and the boilerplate components in the template in sync.  I'm not exactly sure what the right approach is, though.  Let's keep thinking about it.

PR #332, opened 2017-11-13
Fixed a bug preventing navigation via omnibar due to loading outside Angular

Blackbaud-PaulCrowder:
Fixes https://github.com/blackbaud/skyux2/issues/1286

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -156,11 +156,18 @@ export class AppComponent implements OnInit {
     }
 
     nav.beforeNavCallback = (item: BBOmnibarNavigationItem) => {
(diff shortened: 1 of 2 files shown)


src/app/app.component.ts
```diff
@@ -156,11 +156,18 @@ export class AppComponent implements OnInit {
     }
 
     nav.beforeNavCallback = (item: BBOmnibarNavigationItem) => {
+      console.log(baseUrl);
```
Blackbaud-BobbyEarl: Debug remnants?

Blackbaud-PaulCrowder: Good catch.

PR #338, opened 2017-11-15
Build improvements

Blackbaud-SteveBrush:
- Using isolated UglifyJSPlugin, which includes parallelization, https://github.com/webpack-contrib/uglifyjs-webpack-plugin/releases/tag/v1.0.0. (The aliased plugin provided by Webpack only includes version `0.4.x`.) Turns out that minifying the source code takes about 25% of the total build time.
- Targeted folders for various loaders
- Disabled sourcemaps for production. Do we really need them? https://webpack.js.org/guides/build-performance/#source-maps
- Disabled type checking for `ngtools` plugin.
- Removed some imports when auth is set to false.

SKY UX is using these changes with success: https://github.com/blackbaud/skyux2/pull/1294

**Before:**

screen shot 2017-11-15 at 12 27 59 pm

**After:**

screen shot 2017-11-16 at 10 02 47 pm

### DIFF
--- a/cli/build.js
+++ b/cli/build.js
@@ -15,7 +15,7 @@ const runCompiler = require('./utils/run-compiler');
 const tsLinter = require('./utils/ts-linter');
 
 function writeTSConfig() {
-  var config = {
+  const config = {
     'compilerOptions': {
       'target': 'es5',
       'module': 'es2015',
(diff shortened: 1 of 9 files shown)


Blackbaud-SteveBrush:
Closing because it doesn't work in Windows. I'll attempt something else (TBD).

PR #339, opened 2017-11-15
Correctly reading current command in shared Karma config

Blackbaud-BobbyEarl:
Fixes https://github.com/blackbaud/skyux2/issues/1298

### DIFF
--- a/config/karma/shared.karma.conf.js
+++ b/config/karma/shared.karma.conf.js
@@ -14,6 +14,8 @@ function getConfig(config) {
   const path = require('path');
   let testWebpackConfig = require('../webpack/test.webpack.config');
   let remapIstanbul = require('remap-istanbul');
+
+  // See minimist documentation regarding `argv._` https://github.com/substack/minimist
   let skyPagesConfig = require('../sky-pages/sky-pages.config').getSkyPagesConfig(argv._[0]);
 
   // Using __dirname so this file can be extended from other configuration file locations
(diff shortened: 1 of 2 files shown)


Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl It wasn't clear from the issue; what action causes this issue? Just typing `skyux e2e` (for example)?

Blackbaud-BobbyEarl:
Sorry for the confusion @Blackbaud-SteveBrush.  `skyux test` is the command in question.  Doing so is not currently passing the command `test` to `getSkyPagesConfig` method.  This fixes that.

That would mean, if a user had `skyuxconfig.test.json`, it would not have been taken in account for the `SkyPagesConfig` that's passed to the webpack config.

config/karma/shared.karma.conf.js
```diff
@@ -14,7 +14,7 @@ function getConfig(config) {
   const path = require('path');
   let testWebpackConfig = require('../webpack/test.webpack.config');
   let remapIstanbul = require('remap-istanbul');
-  let skyPagesConfig = require('../sky-pages/sky-pages.config').getSkyPagesConfig(argv.command);
+  let skyPagesConfig = require('../sky-pages/sky-pages.config').getSkyPagesConfig(argv._[0]);
```
Blackbaud-SteveBrush: Non-blocking comment regarding readability: I wasn't aware of the underscore property being attached to `argv` (by minimist?) and had to look it up. It'd probably make more sense to assign the value to a human-readable variable, and pass _that_ into the method. Something like:

```
// Some comment describing what `argv._` is...
const cliCommand = argv._[0];
const skyPagesConfig = require('../sky-pages/sky-pages.config').getSkyPagesConfig(cliCommand);
```

PR #340, opened 2017-11-17
Faster build

Blackbaud-SteveBrush:
Turns out the build was pulling from node_modules, which is why it was so slow. There are a few other updates I'd like to make, but this should do for now.

This UX pull request is referencing this branch:
https://github.com/blackbaud/skyux2/pull/1294

### DIFF
--- a/config/webpack/common.webpack.config.js
+++ b/config/webpack/common.webpack.config.js
@@ -75,8 +75,9 @@ function getWebpackConfig(skyPagesConfig, argv = {}) {
       rules: [
         {
           enforce: 'pre',
-          test: /runtime\/config\.ts$/,
-          loader: outPath('loader', 'sky-app-config')
+          test: /config\.ts$/,
+          loader: outPath('loader', 'sky-app-config'),
+          include: outPath('runtime')
         },
         {
           enforce: 'pre',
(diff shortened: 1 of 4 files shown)


Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl Alright, this is finally ready for review.

config/webpack/common.webpack.config.js
```diff
@@ -75,8 +75,9 @@ function getWebpackConfig(skyPagesConfig, argv = {}) {
       rules: [
         {
           enforce: 'pre',
-          test: /runtime\/config\.ts$/,
-          loader: outPath('loader', 'sky-app-config')
+          test: /config\.ts$/,
+          loader: outPath('loader', 'sky-app-config'),
+          include: outPath('runtime')
```
Blackbaud-BobbyEarl: Just curious about the different syntax here.

Blackbaud-SteveBrush: it will ONLY look in the root's runtime folder, instead of "any" runtime folder.

Blackbaud-SteveBrush: Not a huge time saver, but recommended by Webpack docs to be more specific with loaders. https://webpack.js.org/guides/build-performance/#loaders

PR #357, opened 2018-02-03
Exposing the log flag and integrating with better logging plugin

Blackbaud-BobbyEarl:
This PR does a few things...

- Make sure all files use internal logger class, instead of loading `winston` directly.
- Adjusts internal logger class to support the `--logColor` flag.  This will be useful in our VSTS build logs where color is not supported.
- Disables `showLevel` for the `winston` transport.  This means no more `info: ...` before every log entry.
- Using a new webpack logger that support log levels: `none` (custom), `minimal`, `compact`, `expanded`, and `verbose`.  See [simple-progress-webpack-plugin](https://github.com/dominique-mueller/simple-progress-webpack-plugin).  ~Currently pointing to my fork until @dominique-mueller has the opportunity to release.~

@blackbaud-johnly tagging you as this would require entries for the `--logLevel`, `--logColor`, and `logFormat` flags.


config/webpack/common.webpack.config.js
```diff
@@ -48,6 +48,56 @@ function getWebpackConfig(skyPagesConfig, argv = {}) {
...
+    new ProcessExitCode(),
+
+    new OutputKeepAlivePlugin({
+      enabled: argv['output-keep-alive']
+    })
+  ];
+
+  // Supporting a custom logging type of none
+  if (argv.log !== 'none') {
+    const log = skyPagesConfig.runtime.command === 'serve' ? 'compact' : 'expanded';
```
Blackbaud-SteveBrush: Is the intent that `skyux build` should produce the "expanded" logs? I use `skyux build -s` often, and it add a ton of logs to the terminal. Curious if you thought about making the log level default to "compact" and put the option in skyuxconfig.json instead? This would allow control of the log level globally, but could be overwritten in command-specific config files.

Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl Just one comment. I like that all logging goes through a single source now!

config/webpack/common.webpack.config.js
Blackbaud-BobbyEarl: Apologies as that is a bit vague.  `compact` and `expanded` produce the same logging, only difference is compact does it all on a single line.  This is good for local development, but of any "CI" it would not be a good experience.  https://github.com/dominique-mueller/simple-progress-webpack-plugin#compact

I am making the assumption here that if you're running build, you're probably on a CI.  Obviously, you can run build locally, but seemed like the safer default to set.

Allowing the default to be set in skyuxconfig.json might not be a bad idea, as long as the command line overrides it.  This would mean users could get verbose logging on the build machine without altering their build definition, which would be nice.

Blackbaud-BobbyEarl:
@Blackbaud-SteveBrush I believe this is ready for another look.  Since you first approved, I've solidifed on `--logColor`, `--logLevel`, and `--logFormat` as the available options.  I've also migrated the logging code to `@blackbaud/skyux-logger`.

package.json
```diff
@@ -44,7 +44,12 @@
     "@angular/router": "4.3.6",
     "@blackbaud/auth-client": "2.2.0",
     "@blackbaud/skyux-lib-help": "1.1.0",
+    "@blackbaud/skyux-logger": "1.0.0",
```
Blackbaud-SteveBrush: I'm assuming we'll need to update this version once your patch gets in. Happy to approve the pull request once it gets released!

PR #359, opened 2018-02-09
Enable hmr

Blackbaud-BobbyEarl:
Investigating solving https://github.com/blackbaud/skyux2/issues/406.  To test, install this branch and run `skyux serve --hmr` . We could also evaluate this being the default behavior.  Whichever is decided, a docs change would be necessary.

### DIFF
--- a/test/cli-serve.spec.js
+++ b/test/cli-serve.spec.js
@@ -61,12 +61,13 @@ describe('cli serve', () => {
     }, webpackDevServer);
   });
 
-  it('should prepent hrm url to entries if devServer.hot is set', (done) => {
+  it('should prepend hrm url to entries if devServer.hot is set', (done) => {
     const publicPath = '/public-path';
     const url = 'my-url';
     const port = 1234;
 
     const f = '../config/webpack/serve.webpack.config';
+    spyOn(logger, 'info');
     mock(f, {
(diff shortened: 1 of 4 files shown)


test/cli-serve.spec.js
```diff
@@ -61,6 +61,40 @@ describe('cli serve', () => {
     }, webpackDevServer);
   });
 
+  it('should prepent hrm url to entries if devServer.hot is set', (done) => {
```
Blackbaud-SteveBrush: "prepent"

Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl I tested this with SKY UX and it works great! My vote would be to make it the default behavior. If not, it would be nice to have a log or something to indicate that HMR is active.

Blackbaud-BobbyEarl:
@Blackbaud-SteveBrush I added a log when hmr is enabled.  There's also already a log in the browser when HMR is enabled.

@Blackbaud-PaulCrowder any opinion on enabling HMR by default?

Blackbaud-BobbyEarl:
I vote we leave this opt-in.  I'd like to get feedback on usage.  If it works well and developer's like it, I think we can introduce it as the default behavior in the next major release.

PR #361, opened 2018-02-23
Bumping the version of RxJS for support

mattfurness:
Bumping RxJS version so that it matches the version bump in skyux2 (https://github.com/blackbaud/skyux2/pull/1493)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blackbaud/skyux-builder",
-  "version": "1.10.0",
+  "version": "1.11.0",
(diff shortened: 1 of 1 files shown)


Blackbaud-BobbyEarl:
Hey @mattfurness.  Thanks for the PR!  I wanted to give you an quick update.

Unfortunately, the version of RXJS used in builder must explicitly match that used in `@blackbaud/skyux`.  It is a flaw we have a plan to fix in a future breaking change release.  Until then, it's a fairly significant coordinated effort to get the version updated.

PR #362, opened 2018-02-26
switch rxjs dependency to use the compatibilty flag

Blackbaud-StewartStephens:
I need the `rxjs` dependency to use the compatibility flag.  This resolves a conflict with a third-party library which also has a dependency on `rxjs`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "raw-loader": "0.5.1",
     "reflect-metadata": "0.1.10",
     "remap-istanbul": "0.9.5",
-    "rxjs": "^5.4.3",
+    "rxjs": "^5.5.6",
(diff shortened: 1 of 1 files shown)


Blackbaud-BobbyEarl:
Hey @Blackbaud-StewartStephens .  Thanks for the PR!  I wanted to give you an quick update.

Unfortunately, the version of RXJS used in builder must explicitly match that used in `@blackbaud/skyux`.  It is a flaw we have a plan to fix in a future breaking change release.  Until then, it's a fairly significant coordinated effort to get the version updated.

Blackbaud-SteveBrush:
Closing until we figure out a plan of action on this.

PR #363, opened 2018-02-28
[HOLD] Added @blackbaud-internal to list of externals

Blackbaud-SteveBrush:


Blackbaud-BobbyEarl:
Hey @Blackbaud-SteveBrush, can you refresh my memory and provide a little context for why this change would be necessary?  This is necessary for an internally published SKY UX library to not attempt to compile other internally published libraries into its bundle, right?

Blackbaud-SteveBrush:
Addressed with this pull request: https://github.com/blackbaud/skyux-builder/pull/410

PR #366, opened 2018-03-02
Contribute `skyAppLinkExternal`

Blackbaud-MitchellThomas:
We need to navigate to an other SPA but maintain service and environment context. So, we need to sidestep the Angular router but still get the same behavior as `skyAppLink` to maintain the runtime context for the SPA.

- First commit for review on overall approach before writing tests

### DIFF
--- a/runtime/directives/sky-app-link-external.directive.ts
+++ b/runtime/directives/sky-app-link-external.directive.ts
@@ -9,9 +9,23 @@ import { ActivatedRoute, Router, RouterLinkWithHref } from '@angular/router';
 })
 export class SkyAppLinkExternalDirective extends RouterLinkWithHref {
 
+  private _queryParams: { [k: string]: any };
+
   @Input()
   set skyAppLinkExternal(commands: any[] | string) {
     this.routerLink = commands;
+  }
+
+  @Input()
+  set queryParams(params: { [k: string]: any }) {
+    this._queryParams = Object.assign(params, this.skyAppConfig.runtime.params.getAll());
+  }
+
+  get queryParams() {
+    if (!this._queryParams) {
+      this._queryParams = Object.assign({}, this.skyAppConfig.runtime.params.getAll());
+    }
+    return this._queryParams;
   }
 
   constructor(
@@ -22,8 +36,6 @@ export class SkyAppLinkExternalDirective extends RouterLinkWithHref {
     private window: SkyAppWindowRef
   ) {
     super(router, route, new PathLocationStrategy(platformLocation, skyAppConfig.skyux.host.url));
-    this.queryParamsHandling = 'merge';
-    this.queryParams = this.skyAppConfig.runtime.params.getAll();
     if (this.window.nativeWindow.window.name && this.window.nativeWindow.window.name !== '') {
       this.target = this.window.nativeWindow.window.name;
     } else {
--- a/runtime/directives/sky-app-link-external.directive.ts
+++ b/runtime/directives/sky-app-link-external.directive.ts
@@ -1,36 +1,29 @@
-import { Directive, Input, HostListener, OnInit } from '@angular/core';
+import { Directive, Input } from '@angular/core';
+import { PathLocationStrategy, PlatformLocation } from '@angular/common';
 import { SkyAppWindowRef } from '../window-ref';
-import { SkyAppConfig } from '@blackbaud/skyux-builder/runtime';
-import { HttpParams } from '@angular/common/http';
+import { SkyAppConfig } from '../config';
+import { ActivatedRoute, Router, RouterLinkWithHref } from '@angular/router';
 
 @Directive({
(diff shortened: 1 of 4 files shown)


runtime/directives/sky-app-link-external.directive.ts
```diff
@@ -0,0 +1,36 @@
...
+import { SkyAppWindowRef } from '../window-ref';
+import { SkyAppConfig } from '@blackbaud/skyux-builder/runtime';
+import { HttpParams } from '@angular/common/http';
+
+@Directive({
+  selector: '[skyAppLinkExternal]'
+})
+export class SkyAppLinkExternalDirective implements OnInit {
+  @Input() private skyAppLinkExternal: string;
+  private externalSPAUrl: string;
```
Blackbaud-PaulCrowder: To remain consistent with terminology used across the rest of the app, this should be named `externalAppUrl` or simply `externalUrl`.

runtime/directives/sky-app-link-external.directive.ts
```diff
@@ -0,0 +1,36 @@
+import { Directive, Input, HostListener, OnInit } from '@angular/core';
+import { SkyAppWindowRef } from '../window-ref';
+import { SkyAppConfig } from '@blackbaud/skyux-builder/runtime';
+import { HttpParams } from '@angular/common/http';
+
+@Directive({
+  selector: '[skyAppLinkExternal]'
+})
+export class SkyAppLinkExternalDirective implements OnInit {
+  @Input() private skyAppLinkExternal: string;
```
Blackbaud-PaulCrowder: This should be refactored to have a `get`ter and a `set`ter so that the `externalUrl` property can be updated whenever the value changes.  Right now you're only setting the URL when the directive is initialized.

runtime/directives/sky-app-link-external.directive.ts
```diff
@@ -0,0 +1,36 @@
...
+        params = params.append(key, queryParams[key]);
+      }
+    }
+
+    this.externalSPAUrl = `${this.skyAppConfig.skyux.host.url}${urlParts[0]}?${params.toString()}`;
+  }
+
+  @HostListener('click', ['$event'])
+  public onClick(event: MouseEvent) {
+    this.window.nativeWindow.location.href = this.externalSPAUrl;
```
Blackbaud-PaulCrowder: There's no need to handle the element's `click` event; simply update the element's `href` property which you can do using a `@HostBinding()` property.  See Angular's implementation of `routerLink` for an example.

https://github.com/angular/angular/blob/5.2.7/packages/router/src/directives/router_link.ts#L182
https://github.com/angular/angular/blob/5.2.7/packages/router/src/directives/router_link.ts#L233

runtime/directives/sky-app-link-external.directive.ts
```diff
@@ -0,0 +1,36 @@
+import { Directive, Input, HostListener, OnInit } from '@angular/core';
+import { SkyAppWindowRef } from '../window-ref';
+import { SkyAppConfig } from '@blackbaud/skyux-builder/runtime';
+import { HttpParams } from '@angular/common/http';
+
+@Directive({
+  selector: '[skyAppLinkExternal]'
```
Blackbaud-PaulCrowder: Make the selector more specific to only allow anchor tags (i.e. `a[skyAppLinkExternal]`) until there's a compelling use case for non-anchor elements.

runtime/directives/sky-app-link-external.directive.ts
```diff
@@ -0,0 +1,36 @@
...
+  private externalSPAUrl: string;
+
+  constructor(
+    private skyAppConfig: SkyAppConfig,
+    private window: SkyAppWindowRef
+  ) { }
+
+  public ngOnInit() {
+    const queryParams: any = this.skyAppConfig.runtime.params.getAll();
+    const urlParts = this.skyAppLinkExternal.split('?');
```
Blackbaud-PaulCrowder: This assumes the `skyAppLinkExternal` value is a string, but Angular's `routerLink` (and therefore `skyAppLink`) also allow arrays so that URLs can be built up in parts that are dynamic for consumer convenience.  For instance, this is valid where `part2` is a public property of the component:

`<a [skyAppLink]="['part-1', part2, 'part-3']">Dynamic link</a>`

Blackbaud-MitchellThomas:
@Blackbaud-PaulCrowder, I addressed your feedback and have written tests against the new approach, which leverages most of the url logic behind `RouterLinkWithHref`

Blackbaud-BobbyEarl:
Hey @Blackbaud-MitchellThomas this looks good to me.  Looks like there's a conflict with master now.  Can you get that addressed?  We can then get it approved and merged in.

Blackbaud-BobbyEarl:
@Blackbaud-PaulCrowder Can you confirm the current state meets your "changes requested?"

PR #370, opened 2018-03-12
Added required parameters.

blackbaud-brandonhare:
I'll update the tests, but I wanted to get some feedback on the implementation first. I'll also fix the formatting (I'm using prettier, thought I turned it off).

### DIFF
--- a/runtime/params.ts
+++ b/runtime/params.ts
@@ -65,11 +65,11 @@ export class SkyAppRuntimeConfigParams {
     const urlSearchParams = getUrlSearchParams(url);
 
     // Get uppercase keys.
-    const allowedKeysUC = allowed.map((key: string) => key.toUpperCase());
+    const allowedKeysUC = allowed.map(key => key.toUpperCase());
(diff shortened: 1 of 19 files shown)


runtime/params.ts
```diff
@@ -83,6 +87,21 @@ export class SkyAppRuntimeConfigParams {
...
+   * Are all the required params defined?.
+   * @name hasAllRequiredParams
+   * @returns {array}
+   */
+  public hasAllRequiredParams(): boolean {
+    if (this.requiredParams.length === 0) {
+      return true;
+    }
+
+    return this.requiredParams.some((param: string) => {
```
Blackbaud-PaulCrowder: Based on my understanding of `some()`, this would return `true` if there's at least one required param with a value.  This would return a false positive in the case where there are more than one required parameters but at least one required parameter doesn't have a value.  I see there's not a unit test that tests this scenario, so I'd like to see that test written as well as any necessary changes to this code.

blackbaud-brandonhare: Good catch. Swapped that to `every`, which is actually what we want. I've also added a test for that condition.

PR #375, opened 2018-03-19
Library i18n resources

Blackbaud-SteveBrush:
Depends on: https://github.com/blackbaud/skyux-builder/pull/371

Instead of reading the locale JSON files at runtime (via GET), I'm writing their contents to the Assets class at build time. Then, based on the locale at runtime, the respective set of strings is returned.


Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl I want to increase the runtime coverage before official submission, but do you mind taking a quick look at this to see if it's going to be a viable solution?

cli/utils/prepare-library-package.js
```diff
@@ -18,15 +18,22 @@ function makePackageFileForDist() {
 }
 
 function copyFilesToDist() {
-  fs.copySync(
-    skyPagesConfigUtil.spaPath('README.md'),
-    skyPagesConfigUtil.spaPath('dist', 'README.md')
-  );
+  try {
```
Blackbaud-BobbyEarl: I don't like wrapping this in a try/catch, especially if we're not going to do anything with the caught `err`.

lib/sky-pages-assets-generator.js
```diff
@@ -29,11 +32,40 @@ function getSource() {
...
+      ))
+    );
+
+  const resources = {};
+  files.forEach((filePath) => {
+    const key = filePath.split('resources_')[1].split('.json')[0];
+    const contents = fs.readJSONSync(filePath);
+
+    if (resources[key]) {
+      resources[key] = Object.assign(resources[key], contents);
```
Blackbaud-BobbyEarl: I'm confused by this line.  Why assign it to the target and return it?

cli/utils/prepare-library-package.js
Blackbaud-SteveBrush: I wrapped it in a `try` so that if the library doesn't include these files (readme, changelog, or assets) it will not fail (due to how `fs` throws errors). I thought it was better than writing out "does directory exist" conditions. If you'd rather have those, I can add them.

lib/sky-pages-assets-generator.js
Blackbaud-SteveBrush: This line merges resources that share the same locale (in this case, if the library and the spa both have an `en_US` file, it will store the strings under the same `en_US` key).

cli/utils/prepare-library-package.js
Blackbaud-BobbyEarl: The `try/catch` would make it ambiguous as to whether there was a real problem copying the files versus if they files didn't exist to copy.  I think I would prefer to not use it.  I'll let you decide if it's worth optimizing now we have three almost identical blocks of code.

Blackbaud-SteveBrush: Sounds good; I'll take a second swing at it!

Blackbaud-SteveBrush:
Closing, in favor of https://github.com/blackbaud/skyux-builder/pull/379

PR #379, opened 2018-03-26
Allow libraries to define their own locale files

Blackbaud-SteveBrush:
Addresses: https://github.com/blackbaud/skyux2/issues/1546

NOTE: I had to fix some unit tests that were reaching a bit too far into other files.


config/sky-pages/sky-pages.config.js
```diff
@@ -22,132 +22,137 @@ function readConfig(file) {
   return fs.readJsonSync(file, 'utf8');
 }
 
-module.exports = {
+/**
```
Blackbaud-SteveBrush: There are zero functional changes to this file; I simply adjusted the format so we can use a few of its methods (namely `spaPath`) via destructuring. (See my changes to `common.webpack.config.js` for an example.)

test/sky-pages-module-generator.spec.js
```diff
@@ -1,24 +1,73 @@
 /*jshint jasmine: true, node: true */
 'use strict';
 
-const codegen = require('../utils/codegen-utils');
+const mock = require('mock-require');
```
Blackbaud-SteveBrush: This spec file was affecting the state of other unit tests, so I mocked out its dependencies and made sure those dependencies' unit tests were covered 100%.

Blackbaud-SteveBrush:

config/sky-pages/sky-pages.config.js
Blackbaud-BobbyEarl: Just want to say I appreciate you calling this out.  It definitely helps when reviewing.

Blackbaud-BobbyEarl:
Great job on this @Blackbaud-SteveBrush!

Aside from the destructing we talked about offline the only other thing that stood out to me is the use of the `try/catch`.  That masks whether there was an error because the file doesn't exist, or because of another error.  I'd prefer to instead use the existsSync and let any other fatal error bubble up.

Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl I've since removed the destructured method imports, and removed the `try/catch` blocks for the `fs` operations. Let me know if I missed anything!

PR #382, opened 2018-03-28
Added test utility from SKY UX

Blackbaud-SteveBrush:


runtime/testing/browser/test-utility.ts
```diff
@@ -0,0 +1,23 @@
+export interface SkyAppTestUtilityEventArgs {
```
Blackbaud-PaulCrowder: I think I'd call this `SkyAppTestUtilityEventOptions` instead.  Ending a class name with `EventArgs` is a common pattern for classes used to pass arguments into an event handler, which is not the case here.  I'd also move it to a separate file so we're only exporting one type from this file.

runtime/testing/browser/test-utility.ts
```diff
@@ -0,0 +1,23 @@
...
+
+function getWindow() {
+  return window;
+}
+
+export class SkyAppTestUtility {
+  public static fireDomEvent(
+    element: EventTarget,
+    eventName: string,
+    args?: SkyAppTestUtilityEventArgs
```
Blackbaud-PaulCrowder: Remember to change this parameter name to match the new class name.

runtime/testing/browser/test-utility.ts
```diff
@@ -0,0 +1,23 @@
+export interface SkyAppTestUtilityEventArgs {
+  bubbles?: boolean;
+  cancelable?: boolean;
+}
+
+function getWindow() {
```
Blackbaud-PaulCrowder: I'm not sure I understand the purpose of this function.  I could see if maybe you wanted to mock out the `window` object you'd want to do something like this, but it's not possible with the code as it is.

PR #383, opened 2018-03-28
Add ability to import individual SKY UX modules

Blackbaud-PaulCrowder:


Blackbaud-SteveBrush:
Looks good to me! Only concern I would have is the legacy config that some SPAs will have once we split up the libraries and this feature becomes deprecated. But I suppose there might be other properties in the config that will change too, so not a huge deal.

Blackbaud-SteveBrush:
@Blackbaud-PaulCrowder: Also, it appears that every module combination I tried required the `SkyErrorModule`; not sure if it's worth adding that one in no matter what. See: https://github.com/blackbaud/skyux-builder/blob/35a038fa5a5982d523b6496965371f3b1ef57a4e/lib/sky-pages-route-generator.js#L129

PR #384, opened 2018-03-29
Switched Webpack devtool for serve

Blackbaud-SteveBrush:
Addresses: https://github.com/blackbaud/skyux2/issues/1325


Blackbaud-BobbyEarl:
This solution feels different than the HMR.  I'd rather sourcemaps work for everyone, out of the box, without having to opt-in to a particular `devTool`.

https://webpack.js.org/configuration/devtool/ has some good documentation on the supported properties and their affects.  

In my testing, `inline-cheap-module-source-map` also worked well, but as mentioned in said documentation, is slightly faster but does have some limitations.

I would be in favor of merging this PR as-is.

PR #390, opened 2018-04-04
Fixed query param issue skyAppLink

Blackbaud-MitchellThomas:
https://github.com/blackbaud/skyux2/issues/1603

### DIFF
--- a/runtime/directives/sky-app-link.directive.spec.ts
+++ b/runtime/directives/sky-app-link.directive.spec.ts
@@ -8,12 +8,12 @@ import { SkyAppConfig } from '../config';
 import { SkyAppLinkDirective } from './sky-app-link.directive';
 
 @Component({
-  template: `<a skyAppLink='test'>Test</a>`
+  template: '<a skyAppLink="test">Test</a>'
 })
 class SkyAppLinkTestComponent { }
 
 @Component({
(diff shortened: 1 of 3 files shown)


Blackbaud-MitchellThomas:
https://github.com/blackbaud/skyux2/issues/1603

runtime/directives/sky-app-link.directive.spec.ts
```diff
@@ -8,24 +8,34 @@ import { SkyAppConfig } from '../config';
 import { SkyAppLinkDirective } from './sky-app-link.directive';
 
 @Component({
-  template: `<a skyAppLink="test">Test</a>`
+  template: `<a skyAppLink='test'>Test</a>`
```
Blackbaud-BobbyEarl: No functional difference, but any reason these are single quotes?  Same question for line 16.

Blackbaud-MitchellThomas: There's no particular reason. It's probably just muscle memory typing things in a TS file vs. html. That particular line got changed because I copied a bit of code from the external link directive tests and when I massaged it to fit this context, that single quote change was all that was left.

I prefer the double quotes, so I've updated the PR to clean that up plus some other development remnants in the spec file for both tests.

PR #391, opened 2018-04-04
Added module to runtime index

Blackbaud-SteveBrush:
This will allow folks to import the runtime module via the barrel:

`import { SkyAppRuntimeModule } from '@blackbaud/skyux-builder/runtime'`

### DIFF
--- a/runtime/i18n/host-locale-provider.ts
+++ b/runtime/i18n/host-locale-provider.ts
@@ -7,7 +7,7 @@ import { Observable } from 'rxjs/Observable';
 
 import {
   SkyAppWindowRef
-} from '@blackbaud/skyux-builder/runtime';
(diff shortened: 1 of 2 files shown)


runtime/i18n/host-locale-provider.ts
```diff
@@ -7,7 +7,7 @@ import { Observable } from 'rxjs/Observable';
 
 import {
   SkyAppWindowRef
-} from '@blackbaud/skyux-builder/runtime';
+} from '@blackbaud/skyux-builder/runtime/window-ref';
```
Blackbaud-SteveBrush: This was causing a strange circular reference error (probably because it's pointing to a barrel in its parent directory).

PR #402, opened 2018-04-25
Update skyAppLink to use optional excludeDefaults to getAll.

Blackbaud-BobbyEarl:
Addressing https://github.com/blackbaud/skyux2/issues/1650

### DIFF
--- a/runtime/params.ts
+++ b/runtime/params.ts
@@ -146,17 +146,15 @@ export class SkyAppRuntimeConfigParams {
    * @returns {Object}
    */
   public getAll(excludeDefaults?: boolean): Object {
-    const filter = (key: string) => excludeDefaults ?
-      this.params[key] !== this.defaultParamValues[key] : true;
+    const filteredParams: { [key: string]: string} = {};
 
-    const reduce = (params: { [key: string]: string }, key: string) => {
-      params[key] = this.params[key];
-      return params;
-    };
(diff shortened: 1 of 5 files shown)


Blackbaud-BobbyEarl:
Let me know if this implementation is more inline what what you were thinking @Blackbaud-PaulCrowder.

runtime/params.ts
```diff
@@ -140,10 +142,21 @@ export class SkyAppRuntimeConfigParams {
...
+  public getAll(excludeDefaults?: boolean): Object {
+    const filter = (key: string) => excludeDefaults ?
+      this.params[key] !== this.defaultParamValues[key] : true;
+
+    const reduce = (params: { [key: string]: string }, key: string) => {
+      params[key] = this.params[key];
+      return params;
+    };
+
+    return this.getAllKeys()
```
Blackbaud-PaulCrowder: My suggestion here would be to write a simple loop to build up the object.  It'll cut down on both the amount of code written as well as the number of loops executed (1 vs. 2) and be more legible IMO.

```
const filteredParams: { [key: string]: string} = {};

this.getAllKeys().forEach(key => {
  if (!excludeDefaults || this.params[key] !== this.defaultParamValues[key]) {
    filteredParams[key] = this.params[key];
  }
});

return filteredParams;
```

PR #404, opened 2018-04-30
Moved failed build logic from webpack into respective callers.

Blackbaud-BobbyEarl:
https://github.com/blackbaud/skyux2/issues/948


cli/build.js
```diff
@@ -180,7 +182,19 @@ function build(argv, skyPagesConfig, webpack) {
   } else {
     localeAssetsProcessor.prepareLocaleFiles();
     const name = argv.serve ? buildServe : buildCompiler;
-    return name(argv, skyPagesConfig, webpack, isAot);
+
+    return name(argv, skyPagesConfig, webpack, isAot)
+      .catch(err => {
+
+        // Conditionally calling process.exit so it can bubble up if being called from e2e
+        if (!imported) {
```
Blackbaud-SteveBrush: @Blackbaud-BobbyEarl Most of the changes look great. I'm confused what "imported" means in this context?

Blackbaud-BobbyEarl: Valid question.  I use that so that I could conditionally handle logging errors / calling `process.exit`.  

When `skyux build` command is called that `imported` param is `falsy`.  That's b/c we want the errors shown and, more importantly, the process to exit and return a non-zero exit code.

When running `skyux e2e`, that file will `require('./build'), and pass in `true` for that parameter as seen at https://github.com/blackbaud/skyux-builder/pull/404/files#diff-27e8eb2b645e7d3144b02020922373aaR140.  This allows the `e2e` command to not only log the errors, but most importantly, lets it's `catch` handler run.  If not, the e2e wouldn't have an opportunity to perform cleanup when something in build fails.

Blackbaud-SteveBrush: That makes sense.

I'm curious if both `build` and `e2e` should consume the same "bundler", so that we wouldn't have to deal with yet another function argument that's specific only to `e2e`. The bundler could return a promise that's either resolved or rejected.

Would build's linting errors also get rejected properly during `e2e`?
https://github.com/blackbaud/skyux-builder/blob/master/cli/build.js#L179

Blackbaud-BobbyEarl: I like that suggestion, and good catch about my current implementation not handling the linting errors.  Let me see what I can work up.  May do some work w/o updating the unit tests first.

Blackbaud-SteveBrush: Cool! No worries if it's a major refactor

Blackbaud-BobbyEarl:
Took some refactoring, but I liked your suggestion @Blackbaud-SteveBrush.  This is ready for another look.

PR #406, opened 2018-05-10
Support leid

Blackbaud-BobbyEarl:
Related to https://github.com/blackbaud/auth-client/pull/65
Addresses https://github.com/blackbaud/skyux2/issues/1671 and https://github.com/blackbaud/skyux2/issues/1670


runtime/auth-http.ts
```diff
@@ -71,15 +71,18 @@ export class SkyAuthHttp extends Http {
     url: string | Request,
     options?: RequestOptionsArgs
   ): Observable<Response> {
-    let tokenArgs: BBAuthGetTokenArgs;
+    let tokenArgs: BBAuthGetTokenArgs = {};
+    let leId: string = this.getLeId();
```
Blackbaud-PaulCrowder: Should this be `const` instead?  We should probably fix `tokenArgs` as well.

package.json
```diff
@@ -42,7 +42,7 @@
     "@angular/platform-browser": "4.3.6",
     "@angular/platform-browser-dynamic": "4.3.6",
     "@angular/router": "4.3.6",
-    "@blackbaud/auth-client": "2.5.0",
+    "@blackbaud/auth-client": "blackbaud/auth-client#support-leid",
```
Blackbaud-PaulCrowder: Remember to update this to point to 2.6.0.

PR #414, opened 2018-06-15
Navigate post message

Blackbaud-BobbyEarl:
Related to https://github.com/blackbaud/skyux-builder-stache-search/pull/28

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -272,7 +272,7 @@ export class AppComponent implements OnInit, OnDestroy {
     if (this.config.runtime.command === 'e2e') {
       this.windowRef.nativeWindow.addEventListener('message', (event: MessageEvent) => {
         if (event.data.messageType === 'navigate-e2e') {
-          this.router.navigate(event.data.url);
+          this.router.navigateByUrl(event.data.url);
         }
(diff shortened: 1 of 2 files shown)


src/app/app.component.ts
```diff
@@ -269,6 +269,14 @@ export class AppComponent implements OnInit, OnDestroy {
       });
     };
 
+    if (this.config.runtime.command === 'e2e') {
+      this.windowRef.nativeWindow.addEventListener('message', (event: MessageEvent) => {
+        if (event.data.messageType === 'navigate-e2e') {
```
Blackbaud-SteveBrush: If `navigate-e2e` is something we made up, do you think it would be a good idea to prefix it with `sky` so that consumers know exactly where it came from (e.g., it's not a Protractor thing)? `skyux-e2e-navigate`...

Blackbaud-BobbyEarl: That's correct.  Here and https://github.com/blackbaud/skyux-builder-stache-search/pull/28/files#diff-1d60131077aaec482bf1054018c2df5bR137 are the only two places that know about the arbitrary command.  I'm happy to add that prefix.  :-)

PR #421, opened 2018-06-20
RC > Config params as an object; always decode URL params

Blackbaud-SteveBrush:
See:
https://github.com/blackbaud/skyux2/projects/9#card-10723415
https://github.com/blackbaud/skyux2/projects/9#card-10723731

### DIFF
--- a/runtime/config-params.ts
+++ b/runtime/config-params.ts
@@ -1,1 +1,6 @@
-export type SkyuxConfigParams = string[] | { [key: string]: boolean | { value?: any; required?: boolean } };
+export type SkyuxConfigParams = {
+  [key: string]: boolean | {
+    value?: any;
+    required?: boolean
+  }
+};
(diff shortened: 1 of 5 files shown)


runtime/config-params.ts
```diff
@@ -1 +1,6 @@
-export type SkyuxConfigParams = string[] | { [key: string]: boolean | { value?: any; required?: boolean } };
```
Blackbaud-BobbyEarl: I'm not necessarily opposed, just don't recall talking explicitly about it.  Was the intention here to remove support for an array of strings?

Blackbaud-SteveBrush: @Blackbaud-BobbyEarl I may've misunderstood the issue, but I took it to mean that `SkyConfigParams` would always be considered an object. This would make it easier to document and support, instead of having two different types.

https://github.com/blackbaud/skyux2/projects/9#card-10723415

Blackbaud-BobbyEarl: No, you're right.  That's a good catch regarding deprecating that syntax.  Sorry for the confusion.

PR #428, opened 2018-06-26
RC > Allow SPAs to bundle stylesheets that live outside of `src/app`

Blackbaud-SteveBrush:
Resolves: https://github.com/blackbaud/skyux2/issues/413

**Example skyuxconfig.json:**
```
{
  "mode": "easy",
  "compileMode": "aot",
  "app": {
    "styles": [
      "@blackbaud/skyux/dist/css/sky.css",
      "src/styles/custom.css"
    ]
  }
}
```

### DIFF
--- a/lib/sky-pages-module-generator.js
+++ b/lib/sky-pages-module-generator.js
@@ -15,6 +15,11 @@ const routeGenerator = require('./sky-pages-route-generator');
  */
 function getSource(skyAppConfig) {
 
+  // Generate style `require()` statements.
+  const styles = skyAppConfig.skyux.app.styles && skyAppConfig.skyux.app.styles
+    .filter(path => !/^http/.test(path))
+    .map(path => `require('style-loader!${path}');\n`);
+
   // Generate these first so we can check for 404 route
   const components = componentGenerator.getComponents(skyAppConfig);
   const componentNames = components.names;
@@ -188,6 +193,8 @@ const routes: Routes = ${routes.declarations};
 const routing = RouterModule.forRoot(routes, { useHash: ${useHashRouting} });
 
 ${enableProdMode}
+
+${styles}
 
 @NgModule({
   declarations: [
--- a/lib/sky-pages-module-generator.js
+++ b/lib/sky-pages-module-generator.js
@@ -18,9 +18,9 @@ const routeGenerator = require('./sky-pages-route-generator');
  */
 function getSource(skyAppConfig) {
   const cssFilePaths = get(skyAppConfig, 'skyux.app.styles', []);
+  const numFiles = cssFilePaths.length;
 
   let styles = '';
-  const numFiles = cssFilePaths.length;
 
   if (numFiles > 0) {
     let hasExternal = false;
--- a/lib/sky-pages-module-generator.js
+++ b/lib/sky-pages-module-generator.js
@@ -16,30 +16,32 @@ const routeGenerator = require('./sky-pages-route-generator');
  * @returns {string} source
  */
 function getSource(skyAppConfig) {
-
-  // Generate style `require()` statements.
   const cssFilePaths = skyAppConfig.skyux.app &&
     skyAppConfig.skyux.app.styles &&
     skyAppConfig.skyux.app.styles;
 
   let styles = '';
+
   if (cssFilePaths) {
-    styles = cssFilePaths
-      .filter(path => {
-        const isExternal = (/^http/.test(path));
+    let hasExternal = false;
 
-        if (isExternal) {
-          logger.warn([
-            '[skyuxconfig.json] External style sheets are not permitted in the `app.styles` array.',
-            'Consider using `app.externals`:\n',
-            'https://developer.blackbaud.com/skyux/learn/reference/configuration'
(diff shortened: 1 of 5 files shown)


test/sky-pages-module-generator.spec.js
```diff
@@ -358,4 +358,29 @@ BBAuth.mock = true;`);
...
+    const config = {
+      runtime: runtimeUtils.getDefaultRuntime(),
+      skyux: runtimeUtils.getDefaultSkyux()
+    };
+
+    config.skyux.app = {
+      styles: [
+        '@foo/bar/style.scss',
+        'src/styles/custom.css',
+        'https://google.com/styles.css'
```
Blackbaud-BobbyEarl: I see you filtering out styles that start with `http`.  Should we show an error / warning and mention the `externals` property?

Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl This is ready for another look!

lib/sky-pages-module-generator.js
```diff
@@ -15,6 +17,31 @@ const routeGenerator = require('./sky-pages-route-generator');
...
 
+  // Generate style `require()` statements.
+  const cssFilePaths = skyAppConfig.skyux.app &&
+    skyAppConfig.skyux.app.styles &&
+    skyAppConfig.skyux.app.styles;
+
+  let styles = '';
+  if (cssFilePaths) {
+    styles = cssFilePaths
+      .filter(path => {
```
Blackbaud-BobbyEarl: No doubt, for most use-cases this array will be very small.  I definitely like how readable the `map/filter/reduce` pattern is, I'm worried about the performance should the array become large.  In this case, I think the code could be just as readable using a single for loop.  Something like:

```
let styles = '';
let hasExternal = false;

if (cssFilePaths) {
  for (const i = 0, const len = cssFilePaths.length; i < len; i++) {
    if (/^http/.test(cssFilePaths[i])) {
      hasExternal = true;
    } else {
      styles += `require('style-loader!${path}');\n`);
    }    
  }
}

If (hasExternal) {
  logger.warn(...);
}
```

This has a few benefits without sacrificing code readability:

- the warning is only shown once despite the number of externals listed (which you could of course still solve using the other pattern).
- Array is only looped through once

I'm happy to discuss further if you disagree @Blackbaud-SteveBrush.

Blackbaud-BobbyEarl:
This property should also be added to the `$schema` file.

Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl Are my changes closer to what you had in mind?

lib/sky-pages-module-generator.js
```diff
@@ -14,6 +16,33 @@ const routeGenerator = require('./sky-pages-route-generator');
  * @returns {string} source
  */
 function getSource(skyAppConfig) {
+  const cssFilePaths = skyAppConfig.skyux.app &&
+    skyAppConfig.skyux.app.styles &&
+    skyAppConfig.skyux.app.styles;
```
Blackbaud-BobbyEarl: Is this a typo or missing `.length`?  

I've recently started using https://www.npmjs.com/package/lodash.get to handle this exact scenario in a single line.

const get = require('lodash.get');
const cssFilesPaths = get(skyAppConfig, 'skyux.app.styles', []);

Blackbaud-BobbyEarl: You can thank @blackbaud-brandonhare for that tip.

Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl Made the required change. I _was_ required to install a new dependency on `lodash.get`, however. Let me know if that's okay.

lib/sky-pages-module-generator.js
```diff
@@ -14,6 +17,32 @@ const routeGenerator = require('./sky-pages-route-generator');
  * @returns {string} source
  */
 function getSource(skyAppConfig) {
+  const cssFilePaths = get(skyAppConfig, 'skyux.app.styles', []);
+
+  let styles = '';
+  const numFiles = cssFilePaths.length;
```
Blackbaud-BobbyEarl: This is non-blocking, but I would've expected this const to be up against the `cssFilePaths` const, then followed by the `let`.  Just stood out a bit.

skyuxconfig-schema.json
```diff
@@ -104,6 +104,10 @@
               }
             }
           }
+        },
+        "styles": {
+          "description": "An array of CSS or SCSS files to be bundled with the SPA.",
+          "type": "array"
```
Blackbaud-BobbyEarl: I think this also needs the type of items in the array.
```
"items": {
  "type": "string"
}
```

Blackbaud-BobbyEarl:
I know @blackbaud-johnly is aware, but tagging him here just as a notification.  This would require a documentation entry; however, it's important to note that this work is going in to our `rc` branch for builder.  It may be easier to create a similar branch for our docs to start getting the  work ready.

blackbaud-johnly:
Created https://github.com/blackbaud/skyux2/issues/1822 to document the new configuration properties.

PR #429, opened 2018-06-27
RC > Adjusted dev dependencies

Blackbaud-SteveBrush:


Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl You approved this a while ago, but I've since changed a few things because it was failing during the e2e step. I would greatly appreciate your eyes on this!

PR #436, opened 2018-07-20
RC > Replaced error component with iframe

Blackbaud-SteveBrush:


Blackbaud-BobbyEarl:
I know I was absent while you and @Blackbaud-PaulCrowder talked about this, but I don't think I'm a fan of this approach as it currently stands.  

I like the efficiency of referencing our own errors SPA, but I see that as problematic for external consumers and for users who want to customize it.  I think I would probably be okay with it, if we gave a better escape hatch.

We do have some logic baked in to not include the `**` route if a `NotFoundComponent` is found, unfortunately, it's not really usable unless you're in "advanced" mode where you define your own routes.

Looking at: https://github.com/blackbaud/skyux-builder/blob/35a038fa5a5982d523b6496965371f3b1ef57a4e/lib/sky-pages-route-generator.js#L128-L136

What if we changed that to still register the route but reference the `NotFoundComponent` that the SPA owner has created, that would be sufficient?

Blackbaud-BobbyEarl:
Maybe something like the following:

```
  let notFoundComponent = {
    componentName: 'NotFoundComponent',
    routePath: ['**'],
    routeParams: []
  }

  if (skyAppConfig.runtime.handle404) {
    notFoundComponent['componentDefinition'] =
      `@Component({ template: '[YOUR-IFRAME-CODE]' }) export class NotFoundComponent { }`;
  }

  entities.push(notFoundComponent);
```

To replace https://github.com/blackbaud/skyux-builder/blob/35a038fa5a5982d523b6496965371f3b1ef57a4e/lib/sky-pages-route-generator.js#L128-L136

Blackbaud-BobbyEarl:
@blackbaud-johnly we should probably take this opportunity to document this functionality, which I was about to do, but I'm having a hard time deciding a good place for it.  Happen to have any suggestions?

Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl @blackbaud-johnly Think we should start a `rc-` branch on the docs repo?

PR #439, opened 2018-07-23
Libraries > Separated transpile and bundle actions

Blackbaud-SteveBrush:
Resolves: https://github.com/blackbaud/skyux2/issues/1851

@Blackbaud-BobbyEarl I know we talked about avoiding `child_spawn`, but in this case I couldn't find a way around it. Library builds are failing because we're attempting to transpile and bundle in the same step, which causes some problems during AoT. To fix it, I'm transpiling _after_ the bundle has been created so that they don't interfere with one another.

To transpile the files, I need to use Angular's `ngc` (for the metadata), but they don't provide an NPM utility that can be used programmatically, from what I gathered. Also, nearly every tutorial I found recommended that we run `ngc` through the command-line. Another critical thing to note: Angular 6 provides a first-class way to generate libraries, so the changes in this pull request should be temporary until we can upgrade Builder/UX to the latest Angular/TypeScript.

Example tutorial: https://medium.com/@cyrilletuzi/how-to-build-and-publish-an-angular-module-7ad19c0b4464

### DIFF
--- a/cli/build-public-library.js
+++ b/cli/build-public-library.js
@@ -73,19 +73,25 @@ function writeTSConfig() {
         ]
       }
     },
-    'exclude': [
-      skyPagesConfigUtil.spaPathTemp('runtime')
+    files: [
+      skyPagesConfigUtil.spaPathTemp('index.ts')
     ]
   };
 
   fs.writeJSONSync(skyPagesConfigUtil.spaPathTemp('tsconfig.json'), config);
 }
 
+/**
+ * Create a "placeholder" module for Angular AoT compiler.
+ * This is needed to avoid breaking changes; in the future,
+ * we should require a module name be provided by the consumer.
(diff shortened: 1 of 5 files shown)


cli/build-public-library.js
```diff
@@ -73,8 +73,8 @@ function writeTSConfig() {
         ]
       }
     },
-    'files': [
-      skyPagesConfigUtil.spaPathTemp('index.ts')
+    'exclude': [
+      skyPagesConfigUtil.spaPathTemp('runtime')
```
Blackbaud-SteveBrush: Do not transpile (or include) files from Builder's runtime directory.

Blackbaud-SteveBrush:
Related: https://github.com/blackbaud/skyux-builder/pull/417

PR #443, opened 2018-07-26
RC > Removed SKY CSS import

Blackbaud-SteveBrush:
- Replaced runtime and karma style loaders with those released in `@skyux/theme`
- Removed all references to `sky.css`.

To test, run against the template's `builder-dev-rc` branch, here:
https://github.com/blackbaud/skyux-template/tree/builder-dev-rc


Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl We're getting a coverage failure because we were never at 100% code coverage for runtime (due to the Pact stuff). I removed a few things from runtime, which is probably why we're getting a different coverage number. Mind doing a sanity check on that?

Blackbaud-SteveBrush:

package.json
```diff
@@ -39,6 +39,7 @@
     "@angular/platform-browser": "^4.3.6",
     "@angular/platform-browser-dynamic": "^4.3.6",
     "@angular/router": "^4.3.6",
+    "@skyux/theme": ">3.0.0-alpha.1",
```
Blackbaud-BobbyEarl: Should this be `>=`?

Blackbaud-SteveBrush: Good catch!

PR #445, opened 2018-07-27
Fixed library build failures on Windows machines

Blackbaud-SteveBrush:
`skyux build-public-library` is not working on Windows machines due to an incorrect usage of `cross-spawn`.

Also, I'm passing up the status code to fail the build if an error is encountered.


Blackbaud-SteveBrush:
@Blackbaud-BobbyEarl Ready for another look.

PR #448, opened 2018-08-03
RC > Instrument different directory for libraries

Blackbaud-SteveBrush:
`skyux test --coverage library`


config/webpack/test.webpack.config.js
```diff
@@ -20,7 +20,13 @@ function getWebpackConfig(skyPagesConfig, argv) {
   const runCoverage = (!argv || argv.coverage !== false);
   skyPagesConfig.runtime.includeRouteModule = false;
   const ENV = process.env.ENV = process.env.NODE_ENV = 'test';
-  const srcPath = path.resolve(process.cwd(), 'src', 'app');
+
```
Blackbaud-BobbyEarl: This is probably more non-blocking feedback, but it stood out to me that we now have to check for `argv` twice.  What are your thoughts on slightly refactoring to something like the following (untested)?

```
  const argvCoverage = argv ? argv.coverage : '';
  const srcPathUnresolved = [process.cwd(), 'src', 'app'];
  const runCoverage = argvCoverage !== false;
  
  if (argvCoverage === 'library') {
      srcPathUnresolved.push('public');
  }

  const srcPath = path.resolve(...srcPathUnresolved);
  skyPagesConfig.runtime.includeRouteModule = false;
  const ENV = process.env.ENV = process.env.NODE_ENV = 'test';
```

config/webpack/test.webpack.config.js
```diff
@@ -17,10 +17,16 @@ function outPath() {
...
-  const srcPath = path.resolve(process.cwd(), 'src', 'app');
+  const argvCoverage = (argv) ? argv.coverage : true;
+  const runCoverage = (argvCoverage !== false);
+
+  let srcPath;
+  if (argvCoverage === 'library') {
+    srcPath = path.resolve(process.cwd(), 'src', 'app', 'public');
+  } else {
+    srcPath = path.resolve(process.cwd(), 'src', 'app');
+  }
```
Blackbaud-SteveBrush: @Blackbaud-BobbyEarl I agree about the `argv.coverage` checks! I updated the logic, there. 

For the `srcPath`, I personally think writing out the paths long-hand for each condition is more readable, since the reader doesn't need to parse out the `array.push` behavior for each condition. Also, keeping them separate allows more flexibility if we want to change the paths in the future (for example, if the library's source path wasn't a descendant of the SPA's source path). Just my opinion. Thoughts on that?

PR #452, opened 2018-08-15
Resolve locale resources

Blackbaud-SteveBrush:
**This branch accomplishes the following:**
1. Allow SKY UX libraries to reference `.skypageslocales` resource files statically (see description below)
1. Related to above, we also need to generate resource files during `skyux test`.
1. Finally, I noticed that Karma is not catching browser errors. I added the appropriate hook for that.

**Why allow SKY UX libraries to reference resource files statically?**

SKY UX 2 libraries will need this (in particular) to avoid breaking changes with how they're currently handling resources. `SkyResources` is used internally by SKY UX components, but it is synchronous (see: https://github.com/blackbaud/skyux2/blob/master/src/modules/resources/resources.ts#L2).

However, the new libraries need to use Builder's `SkyAppResources` service, which is asynchronous.

This pull request will allow libraries to reference the static files generated by Builder in a _synchronous_ way, until the libraries can release a breaking change to accommodate the asynchronous behavior of the service.

```
// Synchronous retrieval:
private resources = require('!json-loader!.skypageslocales/resources_en_US.json');

public getString(key: string): string {
  return this.resources[key].message;
}
```

Example, being used by `@skyux/core`:
https://github.com/blackbaud/skyux-core/pull/4/files#diff-2cae0aabc5da219240008c726619282eR24

### DIFF
--- a/cli/build-public-library.js
+++ b/cli/build-public-library.js
@@ -70,6 +70,9 @@ function writeTSConfig() {
       'paths': {
         '@blackbaud/skyux-builder/*': [
           '*'
+        ],
+        '.skypageslocales/*': [
+          '../src/assets/locales/*'
         ]
       }
     },
--- a/config/webpack/test.webpack.config.js
+++ b/config/webpack/test.webpack.config.js
@@ -101,6 +101,10 @@ function getWebpackConfig(skyPagesConfig, argv) {
         {
           test: /\.html$/,
           loader: 'raw-loader'
+        },
+        {
+          test: /\.json$/,
+          loader: 'json-loader'
         }
       ]
     },
--- a/config/webpack/test.webpack.config.js
+++ b/config/webpack/test.webpack.config.js
@@ -101,10 +101,6 @@ function getWebpackConfig(skyPagesConfig, argv) {
         {
           test: /\.html$/,
           loader: 'raw-loader'
-        },
-        {
-          test: /\.json$/,
-          loader: 'json-loader'
         }
       ]
     },
--- a/lib/locale-assets-processor.js
+++ b/lib/locale-assets-processor.js
@@ -11,19 +11,26 @@ const localesPath = ['src', 'assets', 'locales'];
 const defaultLocaleFileName = 'resources_en_US.json';
 const defaultFile = tempPath(defaultLocaleFileName);
 
-const libPath = spaPath(
-  'node_modules',
-  '@blackbaud',
-  '**',
-  ...localesPath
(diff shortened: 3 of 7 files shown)


lib/locale-assets-processor.js
```diff
@@ -11,19 +11,26 @@ const localesPath = ['src', 'assets', 'locales'];
...
-  ...localesPath
-);
-
-const libInternalPath = spaPath(
-  'node_modules',
-  '@blackbaud-internal',
-  '**',
-  ...localesPath
-);
+const libPaths = [
```
Blackbaud-SteveBrush: Changing this to an array so that we can add any number of package names.

cli/build-public-library.js
```diff
@@ -70,6 +70,9 @@ function writeTSConfig() {
       'paths': {
         '@blackbaud/skyux-builder/*': [
           '*'
+        ],
+        '.skypageslocales/*': [
```
Blackbaud-SteveBrush: Allow libraries to reference `.skypageslocales` during `skyux build-public-library`.

tsconfig.json
```diff
@@ -26,6 +26,9 @@
       "@blackbaud/skyux-builder/*": [
         "./node_modules/@blackbaud/skyux-builder/*",
         "./*"
+      ],
+      ".skypageslocales/*": [
```
Blackbaud-SteveBrush: Allow libraries to reference `.skypageslocales` during `skyux serve` and `skyux build`.

config/webpack/test.webpack.config.js
```diff
@@ -101,6 +101,10 @@ function getWebpackConfig(skyPagesConfig, argv) {
         {
           test: /\.html$/,
           loader: 'raw-loader'
+        },
+        {
+          test: /\.json$/,
+          loader: 'json-loader'
```
Blackbaud-SteveBrush: This will allow JSON file processing during skyux test.

lib/locale-assets-processor.js
```diff
@@ -11,19 +11,26 @@ const localesPath = ['src', 'assets', 'locales'];
...
+  ),
+  spaPath(
+    'node_modules',
+    '@blackbaud-internal',
+    '**',
+    ...localesPath
+  ),
+  spaPath(
+    'node_modules',
+    '@skyux',
```
Blackbaud-SteveBrush: ^ Will support our new libraries.

PR #458, opened 2018-08-23
Adopt @skyux/theme

Blackbaud-SteveBrush:


package.json
```diff
@@ -50,6 +50,7 @@
     "@pact-foundation/pact": "5.3.0",
     "@pact-foundation/pact-node": "6.7.0",
     "@pact-foundation/pact-web": "5.3.0",
+    "@skyux/theme": "3.0.0-alpha.9",
```
Blackbaud-SteveBrush: This was added as a dependency to make the tests pass, but it should be removed once it's added as a dependency of SKY UX.
See: https://github.com/blackbaud/skyux2/pull/1930

PR #461, opened 2018-08-29
Updated bootstrapper to resolve required legal entity context

Blackbaud-PaulCrowder:


runtime/bootstrapper.ts
```diff
@@ -23,35 +23,36 @@ export class SkyAppBootstrapper {
...
-                '',
-                addQSParam(url, 'envid', envId)
-              );
-            }
-          });
+            url: currentUrl
+          };
+
+          return BBContextProvider.ensureContext(ensureContextArgs)
+            .then((args) => {
```
Blackbaud-BobbyEarl: Does `args` have a type?

Blackbaud-PaulCrowder: It does, but because `BBContextProvider.ensureContext()` has an explicit return type of `Promise<BBContextArgs>`, the type for the `args` parameter in the callback is implicitly typed `BBContextArgs`.  Therefore there's no need to specify the type in the code.

PR #464, opened 2018-09-05
Fix lint/tsc issues when the strict tsc flag is true

mattfurness:
Recently the Stacktrace team have found issues in their spas/libs that could have been caught earlier by the tsc if `strict` was turned on, particularly `strictNullChecks`. If we try to turn on `strict` in our spas/libs we get errors that originate in skyux-builder because the builder is built in to the spas/libs. This PR aims to do the least amount of work possible to allow consumers to turn on `strict` if they choose.

It fixes the linting issues without changing any functionality or public / private api of skyux builder, mostly by:
- Changing the type with the issues into a union of `undefined`
- Using the [Non-null assertion operator](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator)

### DIFF
--- a/runtime/bootstrapper.ts
+++ b/runtime/bootstrapper.ts
@@ -28,7 +28,7 @@ export class SkyAppBootstrapper {
 
           const params = new SkyAppRuntimeConfigParams(
             currentUrl,
-            this.config.params
+            this.config.params!
           );
 
           const ensureContextArgs: BBContextArgs = {


Blackbaud-PaulCrowder:
@Blackbaud-BobbyEarl I took a look and I think I'm fine with these changes, but I'd like to get your take on it as well.

Blackbaud-BobbyEarl:
Hey @mattfurness , can you clarify how I might best see these errors prior to your changes?  I ran `skyux new`, updated the `tsconfig.json` file to include `strict` and `strictNullChecks` in the `compilerOptions` section.  I do get a couple errors, but none that would seem addressed by your PR.

mattfurness:
The errors I found and fixed here were from turning on strict temporarily in this repo and running the tests. I have created a temporary PR in the internal BB `sky-lib-p2p-events` repo that shows around 11 errors from builder in the failed`skyux test` step.  My hunch is that you need to import from `@blackbaud/skyux-builder/runtime` explicitly to start seeing the errors.

bentefay:
Hi y'all. I work with @mattfurness. We noticed that you have updated a bunch of the files in `@blackbaud/skyux-builder/runtime` to import from other precompiled npm packages. Fantastic work! This has resolved most of the compile errors that result from setting `strict: true`.

We have updated our PR to fix the one remaining error. You can reproduce this yourself by running `skyux new`, setting `compilerOptions": { "strict": true }` in `tsconfig.json`, and then updating `HomeComponent` to import `SkyuxConfig` as follows:

```
import { Component } from '@angular/core';
import { SkyAppConfig } from '@blackbaud/skyux-builder/runtime';

@Component({
  selector: 'my-home',
})
export class HomeComponent {
  constructor(config: SkyAppConfig) {
    console.log(config);
  }
}
```

We'd really appreciate you accepting this PR, as `strict: true` will *significantly* improve our ability to handle `null` and `undefined` properly.

Blackbaud-BobbyEarl:
@mattfurness and @bentefay.  First off, my sincerest apologies for being unresponsive on this issue.

Thanks for hanging with us.  This LGTM, but will look for additional validation from @Blackbaud-PaulCrowder and @Blackbaud-SteveBrush.

PR #465, opened 2018-09-05
Disable aria allowed attr

blackbaud-conorwright:
This should remain disabled until we can upgrade to using axe-core 3.1.1. The current version does not recognize certain attributes as appropriate for certain roles that we use together and will cause failures to this rule.

### DIFF
--- a/config/axe/axe.config.js
+++ b/config/axe/axe.config.js
@@ -50,7 +50,7 @@ const defaults = {
     'bypass': { 'enabled': true },
     'tabindex': { 'enabled': true },
 
-    'aria-allowed-attr': { 'enabled': true },
+    'aria-allowed-attr': { 'enabled': false },
     'aria-required-attr': { 'enabled': true },
     'aria-required-children': { 'enabled': true },
     'aria-required-parent': { 'enabled': true },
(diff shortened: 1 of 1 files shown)


config/axe/axe.config.js
```diff
@@ -50,7 +50,8 @@ const defaults = {
     'bypass': { 'enabled': true },
     'tabindex': { 'enabled': true },
 
-    'aria-allowed-attr': { 'enabled': true },
+    // TEMP: this should be re-enabled when we upgrade to axe-core ^3.1.1
```
Blackbaud-SteveBrush: We usually prefix comments like this with `// TODO:` so that they are easily searched for later on.

Also, could you:
1. Put a link in the comment to the Stack Overflow (or similar) article describing the problem?
2. Create an issue in Builder so that we don't forget to take care of this in version 2.

blackbaud-conorwright: Issue made: https://github.com/blackbaud/skyux2/issues/1964
Updated comment

Blackbaud-SteveBrush:
Related issue: https://github.com/blackbaud/skyux2/issues/1964

PR #469, opened 2018-09-12
Added plugins hook to library bundler

Blackbaud-SteveBrush:
- This will allow plugins to affect the `skyux build-public-library` phase.
- Needed to accommodate the synchronous nature of `SkyResources` service.

This is being used, here: https://github.com/blackbaud/skyux-core/pull/31

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,7 @@ environment:
   nodejs_version: "6"
 
 install:
-  - choco install googlechrome
+  - choco install googlechrome --ignore-checksums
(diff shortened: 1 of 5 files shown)


appveyor.yml
```diff
@@ -6,7 +6,7 @@ environment:
   nodejs_version: "6"
 
 install:
-  - choco install googlechrome
+  - choco install googlechrome --ignore-checksums
```
Blackbaud-SteveBrush: See: https://github.com/appveyor/ci/issues/2089#issuecomment-367959300

PR #486, opened 2018-10-16
Added host locale provider

Blackbaud-SteveBrush:


runtime/i18n/host-locale-provider.ts
```diff
@@ -1,3 +1,48 @@
...
+  SkyAppLocaleProvider
+} from './locale-provider';
+
+@Injectable()
+export class SkyAppHostLocaleProvider extends SkyAppLocaleProvider {
+
+  public static readonly defaultLocale = 'en-US';
+
+  constructor(
+    private windowRef: SkyAppWindowRef
```
Blackbaud-SteveBrush: FYI: I had to remove `SkyAppLocaleProvider` from the constructor because Angular was barking about a circular injection. See: https://github.com/blackbaud/skyux-i18n/blob/master/src/app/public/modules/i18n/host-locale-provider.ts#L29

PR #487, opened 2018-10-18
Fix lint/tsc issues when the strict tsc flag is true (again)

bentefay:
Hey guys, thanks for getting [#464](https://github.com/blackbaud/skyux-builder/pull/464) merged and released! Unfortunately, the additional commit that was included in the `1.25` release ([d4c34b7](https://github.com/blackbaud/skyux-builder/commit/d4c34b7eb719fb5c0c9b33abc8a3e19d71cc48ef)) happened to introduce a new TSLint error, which means we still can't turn on Typescript's `strict` mode. This PR fixes that error.

Obviously this approach of us post-hoc fixing `strict` mode compile errors is not really sustainable long term. However, in the short term, we'd really like to get at least one release of `skyux-builder` where we can switch `strict` mode on. Hence this PR. If you could possibly create a patch release with just this commit, it would be fantastic! We can then stay on that version until we can come up with a more sustainable solution. For the longer term, I'm happy to create an issue to discuss some possible permanent solutions, but just as an example we could: 
1. Turn on `strict` mode for the local build of `skyux-builder`. We would have to be careful *not* to change the default `tsconfig.json`, as this is referenced in the `skyux new` template, and we do not want to effect all consuming projects.
2. An even better solution would be to compile the typescript files in `skyux-builder/runtime`. That way typescript files in `skyux-builder` couldn't effect consumers of the library, allowing them to use whatever typescript compiler settings (and ideally, version) that they like.

Any thoughts?

# Additional Info

You can reproduce this error yourself by running `skyux new`, setting `compilerOptions": { "strict": true }` in `tsconfig.json`, and then updating `HomeComponent` to import `SkyuxConfig` as follows:

```
import { Component } from '@angular/core';
import { SkyAppConfig } from '@blackbaud/skyux-builder/runtime';
 
@Component({
  selector: 'my-home',
})
export class HomeComponent {
  constructor(config: SkyAppConfig) {
    console.log(config);
  }
}
```

When running `skyux lint`, you should see the error:
 
```
Error at .../@blackbaud/skyux-builder/runtime/i18n/host-locale-provider.ts:39:14: Variable 'locale' is used before being assigned.
```


jamesottaway:
Thanks Bobby for re-queuing the failed build step. Now that this is green is it possible to merge and release a patch version with this change included today?

PR #500, opened 2018-11-28
Checking if files exists before writing in generate.  Allowing --force/-f to bypass.

Blackbaud-BobbyEarl:
https://github.com/blackbaud/skyux2/issues/1505

### DIFF
--- a/cli/generate.js
+++ b/cli/generate.js
@@ -3,11 +3,24 @@
 
 const fs = require('fs-extra');
 const path = require('path');
+const logger = require('@blackbaud/skyux-logger');
 
-function resolveFilePath(pathParts, fileName) {
+function safeFileWrite(pathParts, fileName, fileContent, force) {
+  const resolvedFileName = path.resolve('src', 'app', ...pathParts, fileName);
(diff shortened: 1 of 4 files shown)


blackbaud-johnly:
Created https://github.com/blackbaud/skyux2/issues/2199 to document the new flags

cli/generate.js
```diff
@@ -60,15 +72,14 @@ function generateComponentTs(pathParts, fileName, name, nameSnakeCase) {
...
 
 }
-`
-  );
-}
+`;
 
-function generateComponentSpec(pathParts, fileName, name, nameSnakeCase) {
-  let nameWithSpaces = properCase(nameSnakeCase.replace(/\-/g, ' '));
+  safeFileWrite(pathParts, fileName + '.ts', fileContent, force);
```
Blackbaud-SteveBrush: Non-blocking, but my branch that adopts eslint will throw an error for this type of string concatenation. It prefers the template:
```
`${fileName}.ts`
```

Blackbaud-BobbyEarl: I went ahead and updated to that style.  :-)

PR #506, opened 2019-01-03
RC > Replace JSCS/JSHint with ESLint

Blackbaud-SteveBrush:


lib/locale-assets-processor.js
```diff
@@ -95,10 +94,11 @@ function getNonDefaultLocaleFiles(dirname) {
 }
 
 function mergeDefaultLocaleFiles() {
-  const libFiles = libPaths.reduce((accumulator, libPath) =>
-    accumulator.concat(
+  const libFiles = libPaths.reduce((accumulator, libPath) => {
+    return accumulator.concat(
```
Blackbaud-PaulCrowder: Does adding `return` here change the functionality at all?

Blackbaud-PaulCrowder: Never mind.  I just realized you added curly braces around it so that adding `return` makes it equivalent to the previous code.  I'm curious as to what rule the original code broke, though.

Blackbaud-SteveBrush: No, since the previous version is returning the value of `accumulator.concat` implicitly. ESlint was complaining about the format of the arrow function, so writing it out explicitly shouldn't affect the end result.

PR #515, opened 2019-02-21
Provide Angular's `LOCALE_ID`

Blackbaud-SteveBrush:
This pull request requires a change to `@skyux/i18n`.
See: https://github.com/blackbaud/skyux-i18n/pull/22

Purpose: This change will automatically set Angular's `LOCALE_ID` based on what's provided by `SkyAppHostLocaleProvider`. The side effect is that Angular's `DatePipe` (and any other localized utilities) will be automatically provided with the same locale string.

### DIFF
--- a/lib/sky-pages-module-generator.js
+++ b/lib/sky-pages-module-generator.js
@@ -90,11 +90,16 @@ function getSource(skyAppConfig) {
       provide: SkyAppLocaleProvider,
       useClass: SkyAppHostLocaleProvider
     }`,
+    `{
+      provide: LOCALE_ID,
+      deps: [SkyAppLocaleProvider],
+      useFactory: skyLocaleIdFactory
+    }`,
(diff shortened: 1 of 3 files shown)


lib/sky-pages-module-generator.js
```diff
@@ -219,6 +224,10 @@ export function SkyAppConfigFactory(windowRef: SkyAppWindowRef): any {
   return config;
 }
 
+export function skyLocaleIdFactory(localeProvider: SkyAppLocaleProvider) {
+  return localeProvider.currentLocale;
```
Blackbaud-SteveBrush: Unfortunately, Angular requires that the return value of a factory be synchronous. I needed to add a synchronous option within `SkyAppHostLocaleProvider` to accommodate this requirement.

Blackbaud-SteveBrush:
Closing in favor of SDK Builder.

PR #522, opened 2019-03-15
Use tslint defocus

Blackbaud-JasonBodnar:
I've been bit too many times by setting a test to `fit` or a spec to `fdescribe` and forgetting to change them back. This change now causes a tslint error when you forget to change them back.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "ts-helpers": "1.1.2",
     "ts-node": "3.0.4",
     "tslint": "5.2.0",
-    "tslint-jasmine-rules": "^1.5.1",
+    "tslint-jasmine-rules": "1.5.1",
     "typescript": "2.3.2",
     "web-animations-js": "2.2.5",
     "webpack": "2.5.1",
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "ts-helpers": "1.1.2",
     "ts-node": "3.0.4",
     "tslint": "5.2.0",
+    "tslint-jasmine-rules": "^1.5.1",
     "typescript": "2.3.2",
     "web-animations-js": "2.2.5",
     "webpack": "2.5.1",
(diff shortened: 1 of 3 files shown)


tslint.json
```diff
@@ -148,6 +149,7 @@
     "use-pipe-transform-interface": true,
     "pipe-naming": [true, "camelCase", "sky"],
     "component-class-suffix": true,
-    "directive-class-suffix": true
+    "directive-class-suffix": true,
+    "no-focused-tests": true
```
Blackbaud-ColbyWhite: it looks like `tslint-jasmine-rules` can do more than the `no-focused-tests`. do we want the others?

```json
{
  "no-focused-tests": true,
  "no-disabled-tests": true,
  "expect-length": true
}
```

Blackbaud-JasonBodnar: I thought about the `no-disabled-tests` but I know in Java land we sometimes have to disable a test (often pact) for some time. `tslint` allows you configure a rule to be on but only a warning. We could do that for `no-disabled-tests`. I haven't looked into what `expect-length` does.

package.json
```diff
@@ -104,6 +104,7 @@
     "ts-helpers": "1.1.2",
     "ts-node": "3.0.4",
     "tslint": "5.2.0",
+    "tslint-jasmine-rules": "^1.5.1",
```
Blackbaud-SteveBrush: Think we could remove the caret `^` and find the latest minor version that works?

tslint.json
Blackbaud-SteveBrush: My vote is to isolate this particular pull request to `no-focused-tests`, since I know of a few folks using disabled tests to keep a running "backlog" of tests they need to write, but haven't yet.
```
xit('should write this test when I have time', () => {});
```

Blackbaud-JasonBodnar: I added `no-disabled-tests` as a warning. I'm assuming that warnings don't prevent things from building, releasing, etc. I tried `skyux test`, `skyux lint`, `skyux server` and `skyux watch` with the warning turned on and everything worked. If does cause problems I'm happy to remove it.

package.json
Blackbaud-JasonBodnar: Caret is removed and `package.json` specifies `1.5.1` which is the latest version.

PR #523, opened 2019-03-19
Update changelog/package.json for 1.33.0 release

Blackbaud-SteveBrush:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.33.0 (2019-03-19)
+
+- Added support for `@skyux/i18n@3.5.1` and `@blackbaud/auth-client@2.13.0`.
```
blackbaud-johnly: I think we should have "[#523](https://github.com/blackbaud/skyux-builder/pull/523)" at the end. I'd have to do some digging to find an example where we link to the current PR like that, but I know I've seen it. And within the changelog, I think we want the link the PR even if the changes are within the same PR as the changelog update.

PR #526, opened 2019-04-29
Update changelog/package.json for 1.34.0 release

Blackbaud-SteveBrush:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.34.0 (2019-04-29)
+
+- Added support for `@skyux/i18n@3.6.0` and `@skyux/router@3.1.1`.
```
blackbaud-johnly: Shouldn't this have a link at the end to this PR? Looks like we are missing it on the previous changelog entry as well, but the entries before that include a link to the PR.

